PR #71, opened 2017-10-12
several ldap server, refs #25915

mzurbuchen:


test/unit/migrations/multiple_ldap_hostnames_migration_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    test 'changes type' do
+      ldap_hostname_setting = Setting.create(id: 8, key: 'ldap_hostname', value: ['yourdomain.com'], type: 'Setting::Host')
+
+      MultipleLdapHostnames.any_instance.expects(:hostname)
+        .times(2)
+        .returns('yourdomain.com')
+
+    Setting.expects(:create)
```
psunix: indention!

test/unit/migrations/multiple_ldap_hostnames_migration_test.rb
```diff
@@ -0,0 +1,104 @@
...
+require migration_file_name
+
+class MultipleLdapHostnamesMigrationTest < ActiveSupport::TestCase
+
+  describe '#up' do
+    test 'does not change type if hostname is an array' do
+      MultipleLdapHostnames.any_instance.expects(:hostname)
+               .returns(['yourdomain.com'])
+
+      assert_nil migration.up
```
psunix: do not rely on return value of up method. test behavior with expects for example.

test/unit/migrations/multiple_ldap_hostnames_migration_test.rb
```diff
@@ -0,0 +1,104 @@
+# encoding: utf-8
```
psunix: never rely on return value of up and down. just because ruby returns the last statement doesn't mean you should use it for testing. the return value here is never used in the application. so please check if the respective setting has been updated correctly by loading and checking it's value.

test/unit/migrations/multiple_ldap_hostnames_migration_test.rb
```diff
@@ -0,0 +1,107 @@
...
+      assert_equal 'Setting::Host', ldap_hostname_setting.type
+      assert_equal ['yourdomain.com'], ldap_hostname_setting.value
+    end
+
+    test 'does not change type if hostname not exists' do
+      Setting.create!(key: 'ldap_hostname', value: nil, type: 'Setting::Text')
+
+      migration.up
+
+      assert_equal 'Setting::Text', ldap_hostname_setting.type
```
psunix: after this migration, ldap_hostname_setting should be a Setting::Host in any cases since this is what this new feature requires

test/unit/migrations/multiple_ldap_hostnames_migration_test.rb
```diff
@@ -0,0 +1,107 @@
...
+      migration.up
+
+      assert_equal 'Setting::Host', ldap_hostname_setting.type
+      assert_equal [], ldap_hostname_setting.value
+    end
+  end
+
+  describe '#down' do
+    test 'does not change type if hostname is a string' do
+      Setting.create!(key: 'ldap_hostname', value: 'yourdomain.com', type: 'Setting::Text')
```
psunix: that should never be the case when your migration#up was executed before !

test/unit/migrations/multiple_ldap_hostnames_migration_test.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    test 'does not change type if hostname not exists' do
+      Setting.create!(key: 'ldap_hostname', value: nil, type: 'Setting::Host')
+
+      MultipleLdapHostnames.any_instance.expects(:hostname)
+        .returns(nil)
+
+      migration.down
+
+      assert_equal 'Setting::Host', ldap_hostname_setting.type
```
psunix: Should be always Setting::Text after migration#down was executed

PR #72, opened 2017-10-12
implemented keyboard shortcuts, refs #23858

mzurbuchen:


mzurbuchen:
- the changes in logins controller are now in a separate commit
- i shortened the texts
- it works on firefox and chrome

PR #74, opened 2017-10-18
made feature tests for quicksearch, refs #25912

mzurbuchen:


test/features/quicksearch_feature_test.rb
```diff
@@ -30,4 +30,22 @@ class QuickSearchFeatureTest < Capybara::Rails::TestCase
...
+    page.must_have_selector('.account-entry')
+    all('.account-entry')[0].click
+    page.must_have_content('account1')
+  end
+
+  test 'search by get params' do
+    login_as_user(:bob)
+    visit('/search?q=account1')
+
+    page.must_have_selector('.account-entry')
```
psunix: please check if the right account is shown. now any account could be returned and the test wouldn't fail.

PR #89, opened 2017-12-04
included db in rubocop except schema.rb and fixed all errors

mzurbuchen:


psunix:
most of these fixes are already on master branch.

PR #90, opened 2017-12-06
added success message and improved messages

mzurbuchen:


app/controllers/api_controller.rb
```diff
@@ -30,7 +34,7 @@ def team
 
   def messages
     @messages ||=
-      { errors: [], info: [] }
+      { errors: [], infos: [], successes: [] }
```
psunix: use same error types as we use in flash messages: error, notice, alert, success
btw: info or information is uncountable :)

app/assets/javascripts/templates/alert_messages.hbs.erb
```diff
@@ -1,4 +1,4 @@
-{{#each info}}
+{{#each infos}}
```
psunix: make this code more DRY ! (dont repeat yourself)
it's multiple time more or less the same ... only css class changes.
maybe you could just create one handlebar template for a message and provide css class and it's messages as parameters.

rename this file to flash_messages since it's the same messages as we us in backend's flash messages

app/views/layouts/_flashes.html.haml
```diff
@@ -21,5 +21,11 @@
       %span{"aria-hidden" => "true"} &#215;
     = flash[:alert]
 
+- if flash[:success]
```
psunix: maket his code/file DRY too

psunix:
changes are obsolete in the meantime

PR #128, opened 2018-03-12
adjusted username regex, refs #35451

mzurbuchen:


app/controllers/admin/users_controller.rb
```diff
@@ -13,6 +13,8 @@ class Admin::UsersController < ApplicationController
 
   helper_method :update_role
 
+  rescue_from ActiveRecord::RecordInvalid, with: :username_invalid_message
```
psunix: use model validation instead of rescuing in controller.

app/controllers/api/accounts_controller.rb
```diff
@@ -21,7 +21,7 @@ def find_accounts(accounts)
     if query_param.present?
       accounts = finder(accounts, query_param).apply
     elsif tag_param.present?
-      accounts.find_by(tag: tag_param)
+      accounts = accounts.find_by(tag: tag_param)
```
psunix: ? what does this have to do with username validation ?
move to sep. commit / branch.

test/controllers/api/accounts_controller_test.rb
```diff
@@ -111,12 +111,10 @@ class Api::AccountsControllerTest < ActionController::TestCase
     end
 
     test 'returns account for matching tag without cleartext username / password' do
-
       login_as(:bob)
+      get :index, params: { 'tag': 'www.tag.ch' }, xhr: true
```
psunix: try to avoid using existing domain names for test data.
use www.example.com vor example

this should probably also not be part of this pull request/commits. please move

test/integration/create_user_test.rb
```diff
@@ -7,43 +7,74 @@
...
+  test 'admin cannot create user with special chars' do
+    login_as('admin')
+    post admin_users_path, params: { user: {
+      username: 'simon?',
+      password: 'password',
+      admin: 0,
+      givenname: 'Simon',
+      surname: 'Kern'
+    } }
+    assert_equal request.fullpath, admin_users_path
```
psunix: do not redirect to admin users path ....
re-render user edit/new if error. use model validations !!!!

psunix:
requested improvements not implemented, closing.

PR #139, opened 2018-05-27
fixed human user policy and view

mzurbuchen:


app/helpers/admin/user_role/dropdown.rb
```diff
@@ -75,7 +75,7 @@ def own_user?
 
       def button_enabled?
         return false if own_user?
-        if @user.admin? && !@current_user.admin?
+        if (@user.admin? && !@current_user.admin?) || @user.root?
```
psunix: isn't it possible to check that by policy ? ask policy ?
policy(...).permission-name?

app/policies/user/human_policy.rb
```diff
@@ -11,6 +11,16 @@ def update?
     current_user.admin?
   end
 
+  def edit?
+    return false if current_user == user
```
psunix: extract to method own_user?
return false if own_user?

app/policies/user/human_policy.rb
```diff
@@ -33,17 +43,11 @@ def update_role?
   end
 
   def destroy?
-    # if user.ldap?
-    #   unless ldap_connection.exists?(user.username)
-    #     return current_user.admin? || current_user.conf_admin?
-    #   end
-    # else
-    return false if user.root?
+    return false if user.root? || current_user == user
```
psunix: also use own_user? method

PR #150, opened 2018-10-24
Remove base64 encoding 

id101010:
Implement different method to generate a random string of alphanumeric characters to avoid equal signs at the end of the generated passwords.


psunix:
created an improved version of the password generator inspired by your solution. 
https://github.com/puzzle/cryptopus/pull/161

PR #163, opened 2019-02-07
fix ip country data fetch

psunix:
implemented a fix and tests pass now.
please test it on integration first. 
this affects running instances on openshift. with current stable code, the build fails! please test and release this changes as soon as possible.


lib/geo_ip.rb
```diff
@@ -0,0 +1,19 @@
+class GeoIp
```
janissbinder: :heart:

PR #180, opened 2019-12-20
Add "setting" rake tasks

niwo:
Add rake tasks to handle ldap_hostname settings


psunix:
we're going to move these ldap settings to env variables so they can be configured env specific. 
this would make your pull request obsolete. Let us know if this would not solve your problem.

niwo:
Yes, this is perfect.

PR #186, opened 2020-02-12
Feature 52742 update account via api

LBruegger:


test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,72 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+      id: account.id,
+      account: {
+        accountname: "Bob Meyer",
+        tag: "taggy"
+      }
+    }
+    patch :update, params: request_account , xhr: true
+
+    assert_equal account.accountname, Account.find(account.id).accountname
+    assert_equal account.tag, Account.find(account.id).tag
```
Thunderguts13: Maybe also check for status code

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,72 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+  end
+
+  context '#update' do
+    test 'update account with valid json structure' do
+      request.headers['Authorization-User'] = bob.username
+      request.headers['Authorization-Password'] = Base64.encode64('password')
+
+      account = accounts(:account1)
+
+      account.save()
```
Thunderguts13: It's not necessary to save the account

app/controllers/api/api_users_controller.rb
```diff
@@ -46,7 +46,7 @@ def destroy
   private
 
   def api_user
-    @api_user ||= User::Api.find(params[:id])
+    @api_user ||= current_user.api_users.find(params[:id])
```
Thunderguts13: Why was this change necessary

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Version 2.4
 * Upgrade to Ruby 2.5, Rails 5.2.3
 * Fix Geo IP DB for new License
+* Remove Root as Admin
+* Add Update Account via Api
```
psunix: Version 2.5 ?
also update Version file !

app/controllers/api/accounts_controller.rb
```diff
@@ -6,21 +6,32 @@
 #  https://github.com/puzzle/cryptopus.
 
 class Api::AccountsController < ApiController
+  helper_method :team
 
+  # GET /api/api_users
```
psunix: api_users ?

app/controllers/api/accounts_controller.rb
```diff
@@ -6,21 +6,32 @@
...
   def show
     account = Account.find(params[:id])
     authorize account
     account.decrypt(decrypted_team_password(account.group.team))
     render_json account
   end
 
+  # PATCH /api/api_users/:id?Query
+  def update
+    account = Account.find(params[:id])
```
psunix: maybe add a method for getting the account by id ? used in show and update method

```
def account
  Account.find ...
end
```

app/controllers/api/team/api_users_controller.rb
```diff
@@ -7,19 +7,22 @@
...
 class Api::Team::ApiUsersController < ApiController
 
+  # GET /api/api_users/
   def index
     authorize team, :team_member?
     @team_api_users = Team::ApiUser.list(current_user, team)
     render_json @team_api_users
   end
 
+  # POST /api/api_users/new
```
psunix: new ?
it's just POST /api/api_users, isn't it ?

app/controllers/api/teams_controller.rb
```diff
@@ -11,18 +11,21 @@ def self.policy_class
     TeamPolicy
   end
 
+  # GET /api/teams
```
psunix: very nice you updated all this missing documentation :dancing_women:

app/policies/account_policy.rb
```diff
@@ -14,6 +14,14 @@ def show?
     team.teammember?(@user.id)
   end
 
+  def update?
+    team.teammember?(@user.id)
+  end
+
+  def permitted_attributes
+    %i[accountname group_id description username password tag]
```
psunix: I think password and username should never be updated directly !
cleartext_username, cleartext_password
see app/controllers/accounts_controller.rb
the username, password attrs of the model are the encrypted values ! 
very important to make sure this cannot be updated directly by a web request !

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,74 @@ class Api::AccountsControllerTest < ActionController::TestCase
       assert_equal 'test', account['cleartext_username']
       assert_equal 'password', account['cleartext_password']
     end
+
+  end
+
+  context '#update' do
+    test 'update account with valid params structure' do
+      request.headers['Authorization-User'] = bob.username
+      request.headers['Authorization-Password'] = Base64.encode64('password')
```
psunix: maybe add a helper method: set_request_headers and add this auth stuff in there.

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,74 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+
+      account = accounts(:account1)
+
+      request_account = {
```
psunix: rename to account_params

make sure to update the username, password ! that's what really matters here ! it's the most important feature of cryptopus, isn't it ? storing and retrieving credentials :man_dancing:

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,74 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+      request.headers['Authorization-Password'] = Base64.encode64('password')
+
+      account = accounts(:account1)
+
+      account.save()
+
+      request_account = {
+        id: account.id,
+        account: {
+          not_valid_key: "Bob Meyer",
```
psunix: use password, username attrs here !
you wanna make sure no one can update them directly !

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,74 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+
+    end
+
+    test 'do not update account when user not in team' do
+    request.headers['Authorization-User'] = alice.username
+    request.headers['Authorization-Password'] = Base64.encode64('password')
+
+    account = accounts(:account2)
+
+    account.save()
```
psunix: no need to save !
account is loaded from the fixtures !
check rails docu about fixtures ! nice feature !

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,74 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+      assert_equal 200, response.status
+    end
+
+    test 'update account without valid params structure should not occur' do
+      request.headers['Authorization-User'] = bob.username
+      request.headers['Authorization-Password'] = Base64.encode64('password')
+
+      account = accounts(:account1)
+
+      account.save()
```
psunix: noooooooo save !
and never add in ruby () when no arguments

test/controllers/api/accounts_controller_test.rb
```diff
@@ -273,6 +273,74 @@ class Api::AccountsControllerTest < ActionController::TestCase
...
+      request_account = {
+        id: account.id,
+        account: {
+          not_valid_key: "Bob Meyer",
+          not_valid_key_2: "taggy"
+        }
+      }
+      patch :update, params: request_account , xhr: true
+
+      assert_equal account.accountname, Account.find(account.id).accountname
```
psunix: he ?
what do you wanna test exactly here? making sure the value did not change ?
if so ...
`
account.reload

assert_equal account.accountname, 'Foobar'
....
`

it's also fine to just check the real value directly ....

psunix: make sure the check the content of the response ! validate JSON in response

test/policies/account_policy_test.rb
```diff
@@ -113,12 +112,25 @@ def team2
...
+
+  def api_alice
+    alice_private_key = alice.decrypt_private_key('password')
+    plaintext_team_password = team1.decrypt_team_password(alice, alice_private_key)
+
+    api_alice = alice.api_users.create
+    team1.add_user(api_alice, plaintext_team_password)
+    api_alice
   end
 end
```
psunix: looks pretty good ! nice you adjusted those tests ! pretty important

app/controllers/api/accounts_controller.rb
```diff
@@ -6,23 +6,44 @@
...
   def show
-    account = Account.find(params[:id])
     authorize account
     account.decrypt(decrypted_team_password(account.group.team))
     render_json account
   end
 
+  # PATCH /api/api_users/:id?Query
+  def update
+    update_account
```
psunix: so you are updating the account before authorizing it ? 
should be authorization always be the first thing to do ?
what does authorize do exactly ?

app/controllers/api/accounts_controller.rb
```diff
@@ -6,23 +6,44 @@
...
+    authorize account
+    account.update!(permitted_attributes(account))
```
psunix: update! saves the records to the db !
you're triggering update and save !
like this all changes are persisted twice !
check api doku for #save and #update

app/controllers/api/accounts_controller.rb
```diff
@@ -6,23 +6,44 @@
...
+    account.save
+    render_json account
+  end
+
   private
 
+  def update_account
```
psunix: I think it's pretty ok just to do this inside #update ... now sep. method required

test/controllers/api/accounts_controller_test.rb
psunix: use password, username as attr for this test !
make sure nobody can override this attrs directly in the db !

psunix: title for test: "account's password and username attributes cannot be set by params"

test/controllers/api/accounts_controller_test.rb
psunix: check the comments above ! account.reload does the trick. no need to use find and get it by id !

test/policies/account_policy_test.rb
```diff
@@ -65,9 +64,9 @@ class AccountPolicyTest < PolicyTest
     test 'non teammember cannot update account' do
       refute_permit alice, account2, :update?
     end
-    
+
     test 'for team enabled api user cannot update account' do
```
psunix: a team enabled api user should be able to update this account, shouldn't he ?
probably just the wrong user ?!

please think about all possible use cases ! 
write them somewhere down and check if you got all possible cases in here !

app/controllers/api/accounts_controller.rb
LBruegger: The authorize Method is a function provided by the pundit gem -  when the method get's called, pundit looks in the same named policy class account. This, in this case is the account_policy class. 
Pundit therefore look if the user is authorized, if not it wil throw an execption. You should also do other account related authorization things, such as the permitted attributes method which checks the allowed attributes in an Object. In my opinion the authorization should always be done before the model is getting updated, but only if an authorization is necessary.

psunix:
I'm going to create a new pull request

PR #190, opened 2020-03-04
Feature 53039 rubocop update

LBruegger:


app/controllers/application_controller.rb
```diff
@@ -25,6 +25,18 @@ class ApplicationController < ActionController::Base
   # includes a security token
   protect_from_forgery with: :exception
 
+  def logout
```
Thunderguts13: The cop for this is disabled. 
`# This cop does not consider action inheritance, therefore disabled
Rails/LexicallyScopedActionFilter:
  Enabled: false`

app/controllers/admin/users_controller.rb
```diff
@@ -7,12 +7,15 @@
...
-  before_action :redirect_if_ldap_user, only: %i[update edit]
-  before_action :authorize_user_class, only: %i[index new create]
-  before_action :authorize_user, only: %i[update unlock edit]
+  before_action :redirect_if_ldap_user, only: [:update, :edit]
+  before_action :authorize_user_class, only: [:index, :new, :create]
+  before_action :authorize_user, only: [:update, :unlock, :edit]
 
   helper_method :update_role
 
+  def edit
```
Thunderguts13: The cop for this is disabled.
`# This cop does not consider action inheritance, therefore disabled Rails/LexicallyScopedActionFilter: Enabled: false`

app/controllers/admin/settings_controller.rb
```diff
@@ -26,10 +26,10 @@ def update_all
 
   private
 
-  def update_attributes(setting_params)
+  def update(setting_params)
```
psunix: don't call this method just update ....
update is usually used as an controller action for PATCH update.
renamed it to update_settings
no need to add params here ... params[:setting] is available in the whole controller

PR #193, opened 2020-03-09
remove locale from url

LBruegger:


app/controllers/application_controller.rb
```diff
@@ -74,7 +74,7 @@ def prepare_menu
   end
 
   def default_url_options(options = {})
-    { locale: I18n.locale }.merge options
+    { locale: nil }.merge options
```
TheWalkingLeek: This is useless now, isn't it?
If you set the locale to nil, it's the same as not setting it at all. Merging it with options just returns options

config/routes.rb
```diff
@@ -10,7 +10,7 @@
   get 'status/health', to: 'status#health'
   get 'status/readiness', to: 'status#readiness'
 
-  scope "(:locale)", locale: /en|de|fr|zh|ru|ch_be/ do
+  scope "/" do
```
TheWalkingLeek: Is this even needed now?
Can't you just remove the scope completely? Feels like that should have the same effect

app/controllers/application_controller.rb
LBruegger: Sure. I didn't know how relevant this, in our case is.

config/routes.rb
LBruegger: Yes, you're right. It is completely unnecessary.

psunix:
pretty good solution ... and it works :)

some room for improvements:

- create a file in config/routes/legacy.rb (we're going to have some more legacy routes in the future) and add the one-liner from the routes.rb
- rename DefaultRoutingController -> LegacyRoutesController
- create dir app/controllers/legacy_routes
- move RemoveLocale to app/controllers/legacy_routes/locale.rb
- nest Locale inside legacy_routes namespace -> LegacyRoutes::Locale
- create integration test file: test/integration/legacy_routes/locale_test.rb and make sure redirecting is working as expected
- check other comments

app/controllers/default_routing_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
+#  Cryptopus and licensed under the Affero General Public License version 3 or later.
+#  See the COPYING file at the top-level directory or at
+#  https://github.com/puzzle/cryptopus.
+
+class DefaultRoutingController
+
+  def redirect
```
psunix: I think it's completely save to just skip authorization here completely. 
make sure you're testing all cases in the new integration test.
logged in and not logged in user for example

psunix: if the user is not logged in, he's going to be redirected to the login form. we also remember what URL the user tried to access and redirect him after successful login. 
check if this is also possible with the legacy url

app/policies/default_policy.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
psunix: remove policy !

app/utils/remove_locale.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
+#  Cryptopus and licensed under the Affero General Public License version 3 or later.
+#  See the COPYING file at the top-level directory or at
+#  https://github.com/puzzle/cryptopus.
+
+class RemoveLocale
+
+  LOCALES = I18n.available_locales.freeze
```
psunix: LOCALES_REGEX = I18n.available_locales.map do |locale| ... .freeze

app/utils/remove_locale.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def initialize
+    @regex_pattern = build_pattern
+  end
+
+  def prepare_url(url)
```
psunix: remove_locale(url)

just return new url without locale in it

app/utils/remove_locale.rb
```diff
@@ -0,0 +1,37 @@
...
+    prepared_url
+  end
+
+  def changed?
+    @changed
+  end
+
+  private
+
+  def not_equal(url1, url2)
```
psunix: =! already means not_equal :)
always add ? to boolean functions

test/integration/legacy_routes/locale_test.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  test 'redirects to account url without locale' do
+    login_as('bob')
+    accounts_url = team_group_accounts_path(teams(:team1).id, groups(:group1).id)
+    get "/de/teams/#{teams(:team1).id}/groups/#{groups(:group1).id}/accounts"
+
+    assert_redirected_to accounts_url
+  end
+
+  test 'redirects to profile url with locale' do
```
psunix: without locale !

PR #194, opened 2020-03-10
Bug 52744 proxy timeout

Thunderguts13:


app/models/maintenance_tasks/removed_ldap_users.rb
```diff
@@ -32,12 +31,12 @@ def enabled?
   private
 
   def collect_removed_ldap_users
-    User::Human.ldap.collect do |user|
-      user unless ldap_connection.exists?(user.username)
-    end.compact
+    User::Human.ldap.reject do |user|
+      ldap_connection.all_uids.include?([user.username])
```
TheWalkingLeek: Is this a two dimensional array? Is that why you include on an array? Would it make more sense to flatten it somewhere if there's only the username in the array?

lib/ldap_connection.rb
```diff
@@ -178,4 +173,10 @@ def search_for_login(username)
     connection.search(base: basename, filter: filter) { |entry| ldap_entry = entry }
     ldap_entry
   end
+
+  def fetch_all_uids
+    filter = Net::LDAP::Filter.pres('uid')
+    users = connection.search(base: basename, filter: filter, attributes: 'uid')
+    users.map(&:uid)
```
TheWalkingLeek: Nice!

test/unit/models/maintenance_tasks/removed_ldap_users_test.rb
```diff
@@ -54,19 +53,13 @@ class RemovedLdapUsersTest < ActiveSupport::TestCase
...
-      expects(:exists?).
-      with('bob').
-      returns(false)
-
-    LdapConnection.any_instance.
-      expects(:exists?).
-      with('alice').
-      returns(true)
+      expects(:all_uids).
+      returns([['alice']]).
```
TheWalkingLeek: Same here as above

app/models/maintenance_tasks/removed_ldap_users.rb
Thunderguts13: Yes you're right I didn't think of that :smile:

PR #195, opened 2020-03-11
Refactor permitted attrs, refs: #53175

TheWalkingLeek:


app/controllers/admin/users_controller.rb
```diff
@@ -89,4 +88,18 @@ def authorize_user_class
   def authorize_user
     authorize user
   end
+
+  def model_params
```
psunix: in this case here, I think it makes sense to move it to the policy class.
here are the attributes depending on the user's role. so it's the policies task.
just call the policies method in #model_params

PR #203, opened 2020-04-09
Feature 53061 webpacker

Thunderguts13:


app/views/accounts/index.html.haml
```diff
@@ -0,0 +1,18 @@
+-#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
```
psunix, line 1: there's no accounts#index route. list of accounts is shown in groups#show, isn't it ?

app/views/wizard/_form.html.haml
```diff
@@ -3,7 +3,7 @@
 -#  See the COPYING file at the top-level directory or at
 -#  https://github.com/puzzle/cryptopus.
 
-= form_tag({action: 'apply'}, id: 'wizardform') do
+= form_tag(wizard_apply_path, method: 'post', id: 'wizardform') do
```
psunix, line 6: post is probably default, isn't it ?
https://apidock.com/rails/ActionView/Helpers/FormTagHelper/form_tag

PR #208, opened 2020-04-16
Restyle account show page, refs: #53704

TheWalkingLeek:


psunix:
is the password strength meter completely removed now?
shouldn't we add this to the edit form ?

psunix:
image

no space between last input field and Update/Cancel buttons

psunix:
please also hide tag input field in form for the moment ...

app/views/accounts/show.html.haml
```diff
@@ -3,16 +3,14 @@
...
 %div
+  %p.description-text
+    = t '.last_change', time_ago: (time_ago_in_words @account.updated_at)
   %p
-    %strong= t :description
-    = simple_format @account.description
-  %p
-    %strong= t :tag
-    = simple_format @account.tag
+    = h @account.description
```
psunix, line 13: nice, we're escaping html  here. https://api.rubyonrails.org/classes/ERB/Util.html#method-c-h
do we also sanitize this content when adding/editing? should we do this in the future ?

LBruegger:
> is the password strength meter completely removed now?
> shouldn't we add this to the edit form ?

We made a separate Story.

PR #212, opened 2020-04-17
Feature 53464 e to e tests

Thunderguts13:


lib/tasks/spec.rake
```diff
@@ -0,0 +1,60 @@
...
+  Rake::Task['spec:system'].actions.clear
+  namespace :spec do
+    RSpec::Core::RakeTask.new(:system) do |t|
+      t.pattern = './spec/system/**/*_spec.rb'
+      t.rspec_opts = '--tag type:system'
+    end
+
+    namespace :system do
+      desc 'Prepare frontend before running system tests'
+      task :prepare do
```
TheWalkingLeek, line 18: We can either run the ember server like developing locally or inject the js and css files, as we do in the docker build...

PR #216, opened 2020-04-27
Bug 53900 form tag path

lucashabersaat:
Most of the form_tag calls are in views, that are not used anymore. Thus, I went through and deleted what I found deprecated.
Of course risk remains that something was mistakenly deleted that is still in use. 

For instance, I wasn't sure about this folder in views: [admin/maintenance_tasks/removed_ldap_users](https://github.com/puzzle/cryptopus/tree/master/app/views/admin/maintenance_tasks/removed_ldap_users). Is not yet deleted, but couldn't find any usages.

Rspec succeeded, but some tests remain in integration like changing password of a ldap user.


lucashabersaat:
Note, the fix for autologoff which has been directly pushed on stable is aswell included here.

app/views/session/new.html.haml
```diff
@@ -4,7 +4,7 @@
 -#  https://github.com/puzzle/cryptopus.
 
 .login-form
-  = form_tag(session_path, method: :post, class: 'login') do
```
LBruegger, line 7: Why is the method unnecessary?

lucashabersaat: Because method defaults to post.

PR #217, opened 2020-04-28
Bug 53886 update users

lucashabersaat:
The other potential solution of just dropping the url argument in the `form_for` to let rails figure out the path didn't work. Because when creating a new user, rails takes the undefined path `admin_user_humen_path`, instead of `admin_users_path`. This seems because User can be Human and API and `admin/users` is a list of `admin/user_humen` actually.


lucashabersaat:
Note, the fix for autologoff which has been directly pushed on stable is aswell included here.

PR #219, opened 2020-04-28
Bug 53886 update users

lucashabersaat:
Access error on integration for editing a user in admin/users.


app/views/admin/users/edit.html.haml
```diff
@@ -7,7 +7,7 @@
 
 - unless @user.root?
```
lucashabersaat, line 8: Why can't root see the form?

PR #222, opened 2020-04-29
Feature 53401 account new modal

TheWalkingLeek:
What is missing is a full, extensive e2e test. I can take care of that. I think it still makes sense to merge it like this though to not block other stories... :t-rex:


psunix:
image

what about bootstrap styling of form components ?

psunix:
how about responsive layout for small display of modal window ?
new story for this ?

frontend/app/components/account-form.js
```diff
@@ -0,0 +1,90 @@
...
+import { tracked } from "@glimmer/tracking";
+import BaseFormComponent from "./base-form-component";
+import { bind } from "@ember/runloop";
+
+export default class AccountForm extends BaseFormComponent {
+  @service store;
+  @service router;
+
+  @tracked selectedTeam;
+  @tracked allTeams = this.store.findAll("team");
```
psunix, line 15: assignableTeams ?
a term we introcuded in SWEB :)
it's never all teams, is it? just the ones that the user has access to

frontend/app/components/base-form-component.js
```diff
@@ -0,0 +1,45 @@
+import { action } from "@ember/object";
```
psunix, line 1: love it !

frontend/app/components/base-form-component.js
```diff
@@ -0,0 +1,45 @@
...
+import { tracked } from "@glimmer/tracking";
+import { inject as service } from "@ember/service";
+
+export default class BaseFormComponent extends Component {
+  @service intl;
+
+  @tracked
+  record;
+
+  async beforeSubmit() {}
```
psunix, line 12: would it make sense to leave some comments in this file?
maybe with an example when one might use one of those hooks?

frontend/app/templates/components/account-form.hbs
```diff
@@ -0,0 +1,72 @@
...
+              </div>
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-sm">
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-sm-6">
+              <label name="description">{{t "account.description"}}</label>
```
psunix, line 57: description field is ab bit small ... isn't it ?

psunix:
what about translation of field labels ?

PR #226, opened 2020-05-12
Feature 53859 security audit smaller issues

LBruegger:


app/assets/javascripts/modules/login.js.coffee
```diff
@@ -28,7 +28,7 @@ class app.Login
     return
 
   setLocalUsername = ->
```
psunix, line 30: remove function completely !

app/controllers/api_controller.rb
```diff
@@ -88,6 +89,12 @@ def authorize_with_headers
     end
   end
 
+  def set_headers
+    response.headers['Content-Type'] = 'text/json'
+    response.headers['Content-disposition'] = "attachment;\
+    filename='" + params['controller'].to_s + "'.json"
```
psunix, line 95: try to avoid using params for static values.
I would rather use request.env['PATH_INFO']

psunix: add this header checking at least to one controller spec

config/environments/production.rb
```diff
@@ -96,4 +96,6 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  config.force_ssl = true
```
psunix, line 100: remove !
it's already in this file -> line 53

config/initializers/session_store.rb
```diff
@@ -1,9 +1,8 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cache_store, key: '_cryptopus_session', expire_after: 5.minutes
+Rails.application.config.session_store :cache_store, key: '_cryptopus_session', expire_after: 5.minutes, same_site: :lax
```
psunix, line 3: same_site: :strict doesn't work, does it ?

psunix:
maybe you could add an integration test named http_headers_spec.rb with a section 'security'
which checks all those headers.

config/initializers/session_store.rb
LBruegger: It truly does work 😅 yesterday, it didn't

PR #227, opened 2020-05-20
Feature 54050 password strength meter

LBruegger:


frontend/app/templates/components/account-form.hbs
```diff
@@ -43,23 +40,23 @@
 
                   <div class="form-group col-sm-6">
                     <div class="row">
-                      <Input class="form-control " @value={{this.changeset.cleartextPassword}} @name="cleartextPassword" @type="password"/>
+                      <Input class="form-control mb-10" @value={{mut this.changeset.cleartextPassword}} @name="cleartextPassword" @type="password" @key-up={{action "updatePasswordScore"}}/>
+                      {{ember-progress-bar progress=this.passwordScore options=(hash strokeWidth=3 duration=1)}}
```
lucashabersaat: Nice, I like the simple look. 
What about colors? Possible?
A subtle red, orange and green would be cool I think. But not sure if worth the effort...

frontend/app/templates/components/account-form.hbs
```diff
@@ -43,23 +40,23 @@
...
                     </div>
                   </div>
 
                   <div class="form-group col-xs-12 col-sm-5 pull-right">
                     <div class="row">
                       <button class="btn btn-secondary btn-block" {{on "click" this.setRandomPassword}}>{{t "randomPassword"}}</button>
+                      <label>{{this.passwordLabel}}</label>
```
lucashabersaat, line 51: I would place the Label on one line right under the progress bar, instead under the Random Button.

frontend/app/styles/app.css
```diff
@@ -40,3 +40,7 @@ li a {
 #team-power-select {
   height: 69px;
 }
+
+.mb-10 {
```
lucashabersaat, line 44: Give it a clearer name? margin-bottom-10?

frontend/app/routes/accounts/new.js
```diff
@@ -1,5 +1,12 @@
...
+import { getOwner } from '@ember/application';
 
 export default Route.extend({
+  get passwordStrength() {
+    return getOwner(this).lookup('service:password-strength');
+  },
 
+  beforeModel() {
+    return this.passwordStrength.load();
+  }
```
lucashabersaat, line 11: See above.

frontend/app/routes/accounts/edit.js
```diff
@@ -1,6 +1,15 @@
 import Route from '@ember/routing/route';
+import { getOwner } from '@ember/application';
 
 export default Route.extend({
+  get passwordStrength() {
+    return getOwner(this).lookup('service:password-strength');
+  },
+
```
lucashabersaat, line 8: Can you move this into _routes/accounts.js_ ? Not sure if that works.

lucashabersaat:
Another thing. If one sees this bar the first time, and doesn't change the password. It's not really clear, what this bar represents. Maybe add a Label StrengthMeter, or make the background grey to indicate more a progress bar. Something that shows, what this little box tells us.

frontend/app/styles/app.css
LBruegger: We already have margin which is named like this. So i made it like them, to maintain constancy.

frontend/app/templates/components/account-form.hbs
LBruegger: Sure.

frontend/app/templates/components/account-form.hbs
LBruegger: Possible. But you need to reload the ProgressBar everytime you want to change the Color.

frontend/app/components/account-form.js
```diff
@@ -11,20 +11,38 @@ import { isPresent, isNone } from '@ember/utils';
...
-
+  @tracked passwordScore;
+  @tracked passwordLabel;
 
   AccountValidations = AccountValidations;
 
+  @action
+  updatePasswordScore() {
+    if (isPresent(this.changeset.cleartextPassword)){
+      this.passwordStrength.strength(this.changeset.cleartextPassword).then(strength => {
```
psunix, line 28: can't we do this inside the service ?

psunix:
we really need to style this :smile: 
how about something like this ?

image

LBruegger:
I the last commit i fixed everything regarding the password strength meter. The logic for the password check is now available in the service. Also i reworked everything. Now it looks a lot better than my previous version.

But there is still one issue missing. I just didn't know how to open the bootstrap popover to check if the values are appropriate for each step of password strength.

frontend/app/components/account-form.js
```diff
@@ -11,17 +11,53 @@ import { isPresent } from "@ember/utils";
...
+    let calculatedScore = passwordScore * 25;
+    let translation = this.intl.t(this.translations[passwordScore])
+    let color = passwordScore - 1 === -1 ? "none" : this.colors[passwordScore - 1]
+      return '<div class="progress">' +
+        '<div class="progress-bar" role="progressbar" style="width: ' + calculatedScore + '%; background-color: ' + color + ' !important;" aria-valuenow="' + calculatedScore + '" aria-valuemin="0" aria-valuemax="100"></div>' +
+        '</div>' + '<p class="font-weight-bold">' + translation + '</p>';
+  }
+
+  popoverTemplate() {
+    return '<div class="popover"><div class="arrow"></div>'+
```
psunix, line 52: :nauseated_face: so much template code inside a component.

frontend/app/styles/app.css
```diff
@@ -66,3 +74,30 @@ li a {
...
+  background-color: #fc6703 !important;
+}
+.progress-bar-75 {
+  width: 75% !important;
+  background-color: #fcc603 !important;
+}
+.progress-bar-100 {
+  width: 100% !important;
+  background-color: #4dd100 !important;
+}
```
lucashabersaat, line 103: Not very happy with this, but it's safer instead of setting style inline. Wanted to put it in its own stylesheet, but didn't know how to in Ember. Hope that's ok for now.

frontend/app/components/password-strength-meter.js
```diff
@@ -1,3 +1,41 @@
...
+
+  @tracked score = 0;
+
+  constructor() {
+    super(...arguments)
+
+    this.passwordStrength.load()
+  }
+
+  didUpdateAttrs() {
```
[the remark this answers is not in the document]
lucashabersaat: This is a hook, so must keep this name.

PR #233, opened 2020-05-27
Correctly reencrypt data when moving accounts to different team, refs…

TheWalkingLeek:
…: #54102


app/models/item.rb
```diff
@@ -90,7 +90,7 @@ def decrypt_attr(attr, team_password)
   end
 
   def encrypt_file(team_password)
-    return if cleartext_file.blank?
+    return if cleartext_file.force_encoding("ISO-8859-1").blank?
```
lucashabersaat, line 93: Nice.

app/controllers/api/accounts_controller.rb
```diff
@@ -37,6 +37,7 @@ def create
   # PATCH /api/accounts/:id?Query
   def update
     authorize account
+    account_move_handler.move(params_group) if team_changing?
```
psunix, line 40: how about account.group.team_id_changed?
I think this should be available already

app/models/item.rb
```diff
@@ -90,7 +90,7 @@ def decrypt_attr(attr, team_password)
   end
 
   def encrypt_file(team_password)
-    return if cleartext_file.blank?
+    return if cleartext_file.force_encoding('ISO-8859-1').blank?
```
psunix: what the encoding for ?

spec/controllers/api/accounts_controller_spec.rb
```diff
@@ -322,6 +323,75 @@
       expect(response).to have_http_status(200)
     end
 
+    it 'updates account and correctly move items' do
```
psunix, line 326: it 'moves account to other team\'s group' ... ?

spec/controllers/api/accounts_controller_spec.rb
```diff
@@ -322,6 +323,75 @@
...
+      login_as(:bob)
+
+      account = accounts(:account1)
+      new_group = groups(:group2)
```
psunix, line 330: target_group ?

spec/controllers/api/accounts_controller_spec.rb
```diff
@@ -322,6 +323,75 @@
...
+            tag: 'taggy',
+            cleartext_username: 'globi',
+            cleartext_password: 'petzi'
+          },
+          relationships: { group: { data: { id: new_group.id, type: 'groups' } } }
+        }, id: account.id
+      }
+      expect do
+        patch :update, params: account_params, xhr: true
+      end.to raise_error(RuntimeError, 'user is not member of new team')
```
psunix, line 380: shouldn't we return a 403 here ?

app/models/item.rb
TheWalkingLeek: actually just updated that. Was a bug because it tried to use a match on an invalid utf-8 string and thus encountered an ArgumentError.
Replaced blank? with empty? now and it works...

app/controllers/api/accounts_controller.rb
TheWalkingLeek: But the team_id of the group didn't change. We moved the account, not the group.
`account.group.team_id_changed? => false`
`account.group_id_changed? => true`
But just because the group_id changed doesn't mean that the account is in a new team

spec/controllers/api/accounts_controller_spec.rb
TheWalkingLeek: depends how we want to handle this. Someone who's not in the target team can still update the other attributes of the account. But can't move it.

app/controllers/api/accounts_controller.rb
psunix: of course ... didn't think it through completely.

psunix: I think it's pretty ok if we just check if the group has changed. since we're taking care of an eventual team change inside AccountMoveHandler.
if account.group_id_changed? or even account.group_changed?

spec/controllers/api/accounts_controller_spec.rb
psunix: aaah, I see. this exception comes from the AccountMoveHändler. leave it this way!

PR #234, opened 2020-06-02
Feature 54051 fileentry modal

lucashabersaat:


psunix:
when I try to upload a file I'm receiving a server error 500

Completed 500 Internal Server Error in 16ms (ActiveRecord: 0.8ms | Allocations: 6689)

invalid byte sequence in UTF-8 excluded from capture: DSN not set
  
ArgumentError (invalid byte sequence in UTF-8):
  
app/models/file_entry.rb:97:in `encrypt_file'
app/models/file_entry.rb:55:in `encrypt'
app/models/file_entry.rb:40:in `create'
app/controllers/api/file_entries_controller.rb:16:in `build_entry'
app/controllers/crud_controller.rb:14:in `create'

PR #236, opened 2020-06-03
Adjust self remove from team handling, refs: #54305

TheWalkingLeek:


app/serializers/teammember_serializer.rb
```diff
@@ -37,6 +37,10 @@ def admin
     user_object.admin?
   end
 
+  def current_user
+    user_object.id == @instance_options[:current_user_id]
+  end
```
lucashabersaat, line 42: :flags:

frontend/app/components/team-member-configure.js
```diff
@@ -53,19 +56,24 @@ export default class TeamMemberConfigureComponent extends BaseFormComponent {
...
+        });
+        this.loadCandidates();
+      }
     });
   }
 
   @action
   addMember(member) {
-    let team = this.store.peekRecord('team', this.args.teamId)
+    let team = this.store.peekRecord("team", this.args.teamId);
```
lucashabersaat, line 73: What is the policy when to use double-quotes and when single-quotes?

PR #238, opened 2020-06-04
Add sentry to frontend, add env-settings endpoint

Thunderguts13:
Sentry is implemented in Ember. The DSN gets retrieved via env_settings endpoint. Right now the Sentry-DSN is the only Parameter on this endpoint. Are there any other Environment Variables which could be included?

### DIFF
--- a/frontend/app/initializers/env-settings.js
+++ b/frontend/app/initializers/env-settings.js
@@ -1,0 +1,19 @@
+import ENV from "../config/environment";
+
+export function initialize(/* application */) {
+  if (ENV.environment !== "test") {
+    /* eslint-disable no-undef  */
+    $.getJSON({
+      url: "/api/env_settings",
+      async: false,
+      success: function (envSettings) {
+        ENV.sentryDsn = envSettings.sentry;
+      }
+    });
+    /* eslint-enable no-undef  */
+  }
+}
+
+export default {
+  initialize
+};
--- a/frontend/app/initializers/sentry.js
+++ b/frontend/app/initializers/sentry.js
@@ -1,0 +1,18 @@
+import * as Sentry from '@sentry/browser'
+import * as Integrations from '@sentry/integrations';
+import ENV from "../config/environment";
+import { isPresent } from '@ember/utils';
+
+export function initialize() {
+  if (ENV.environment === "production" && isPresent(ENV.sentryDsn)) {
+    Sentry.init({
+      dsn: ENV.sentryDsn,
+      integrations: [new Integrations.Ember()]
+    });
+  }
+}
+
+export default {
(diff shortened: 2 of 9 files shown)


frontend/app/initializers/env-settings.js
```diff
@@ -0,0 +1,17 @@
+import config from "../config/environment";
+
+export function initialize(/* application */) {
+  /* eslint-disable no-undef  */
+  $.getJSON({
+    url: "/api/env_settings",
+    async: false,
```
Thunderguts13, line 7: Sentry relies upon the synchronous call to the Backend to receive the dsn in the right moment. The problem is that this Mehthod is depracated. So I'm open for suggestions for a different solution.

spec/controllers/api/env_settings_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      sentry_env_vars
+
+      get :index
+
+      expect(json['sentry']).to eq('123456')
+    end
+  end
+end
+
+def sentry_env_vars
```
TheWalkingLeek, line 21: maybe rename this to `prepare_sentry_env_vars`? and put it inside a `before(:each)`?

frontend/app/initializers/env-settings.js
TheWalkingLeek: Don't really know a different way of handling this...

frontend/app/initializers/sentry.js
```diff
@@ -0,0 +1,18 @@
+import * as Sentry from '@sentry/browser'
+import * as Integrations from '@sentry/integrations';
+import ENV from "../config/environment";
+import { isPresent } from '@ember/utils';
+
+export function initialize() {
+  if (ENV.environment === "production" && isPresent(ENV.sentryDsn)) {
```
psunix, line 7: nice !

spec/controllers/api/env_settings_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Api::EnvSettingsController do
+  include ControllerHelpers
+
+  before(:each) do
```
psunix, line 8: before (without :each) should do it :)

spec/controllers/api/env_settings_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    ENV['SENTRY_DSN_FRONTEND'] = '123456'
+  end
+
+  context 'GET index' do
+
+    it 'returns env_settings with keycloak enabled' do
```
psunix, line 14: keycloak ?

PR #240, opened 2020-06-05
Improve modal styling, refs: #54302

TheWalkingLeek:
The checkbox inside `/app/views/recryptrequest/new_ldap_password` was not buggy to me... I tried this with firefox and chromium. Is this still happening and if so what browser are you using?

### DIFF
--- a/frontend/app/styles/app.css
+++ b/frontend/app/styles/app.css
@@ -1,3 +1,23 @@
+.ember-power-select-typeahead-trigger {
+  border: none;
+  border-radius: 0;
+}
+
+.ember-power-select-typeahead-input {
+  position: absolute;
+  left: 0;
+  right: 0;
+  width: 100%;
+  top: 0;
+  bottom: 0;
+  border: 1px solid #aaaaaa !default;
+  outline: none;
+}
+
+.ember-power-select-typeahead-input:focus {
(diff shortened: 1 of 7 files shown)


frontend/app/styles/app.css
```diff
@@ -20,8 +40,10 @@ li a {
 
 /* Validation-errors */
 .validation-errors {
-  color: red;
-  font-size: small;
+  width: 100%;
+  margin-top: 0.25rem;
+  font-size: 90%;
```
lucashabersaat, line 45: In my opinion, validation errors look good enough like this. I remember @psunix saying, we might want them to stick out more?

frontend/app/styles/app.css
```diff
@@ -1,3 +1,23 @@
...
+  right: 0;
+  width: 100%;
+  top: 0;
+  bottom: 0;
+  border: 1px solid #aaaaaa !default;
+  outline: none;
+
+  & :focus {
+    border: 1px solid #aaaaaa !default;
+  }
```
lucashabersaat, line 18: Maybe add some label? Placeholder? Text-box alone might be not clear enough.

lucashabersaat:
Concerning the Forgot Password. Well, it's not buggy anymore for me aswell. Anyhow, clicking the label does not check the checkbox.

PR #248, opened 2020-06-10
Fix folder index method, increase tests

Thunderguts13:


app/controllers/api/folders_controller.rb
```diff
@@ -45,7 +38,8 @@ def update
 
   private
 
-  def find_folders(folders)
+  def fetch_entries
+    folders = team.folders
```
TheWalkingLeek, line 42: Does this seem better to you: ?
```
return team.folders unless query_param.present?

finder(team.folders, query_param).apply
```
Just as an idea

PR #249, opened 2020-06-10
Feature 54339 fix api user bug

LBruegger:


app/assets/javascripts/modules/api_user_toggle.js.coffee
```diff
@@ -24,13 +24,13 @@ class app.ApiUserToggle
   lockApiUser = (id) ->
     $.ajax({
       method: 'POST',
-      url: '../api/api_users/' + id + '/lock'
-    })
+      url: '../api/api_users/' + id + '/lock',
```
TheWalkingLeek, line 27: Won't need this `,` right? :)

app/assets/javascripts/modules/api_user_toggle.js.coffee
```diff
@@ -24,13 +24,13 @@ class app.ApiUserToggle
...
+    });
 
   unlockApiUser = (id) ->
     $.ajax({
       method: 'DELETE',
-      url: '../api/api_users/' + id + '/lock'
+      url: '../api/api_users/' + id + '/lock',
```
TheWalkingLeek, line 33: same here

app/assets/javascripts/modules/api_users.js.coffee
```diff
@@ -23,10 +23,7 @@ class app.ApiUsers
     appendRow(row)
 
   apiUsersData = (data) ->
-    apiUsers = data['data']
-    if(!apiUsers)
-      apiUsers = [data['data']]
-    apiUsers
+    data['data']
```
TheWalkingLeek, line 26: You tested this thoroughly right? This could break a couple of things i imagine

LBruegger: Yes, I think it is thoroughly tested. I assume that every test case is covered in this [test](https://github.com/puzzle/cryptopus/blob/feature-54339-fix-api-user-bug/spec/controllers/api/api_users/lock_controller_spec.rb) as well in this [one](https://github.com/puzzle/cryptopus/blob/feature-54339-fix-api-user-bug/spec/controllers/api/api_users_controller_spec.rb). The Problem here was that the JSON structure didn't match the parameter of the [apiUsersContent](https://github.com/puzzle/cryptopus/blob/a0a66d32910a6249a610945e78656a026185c121/app/assets/javascripts/modules/api_users.js.coffee#L17) Method. I fixed this by adding Brackets to the loaded data, when a new ApiUser gets created.

PR #252, opened 2020-06-12
Implement logout for keycloak

Thunderguts13:


app/assets/javascripts/modules/auto_logoff.js.coffee
```diff
@@ -15,7 +15,7 @@ class app.AutoLogoff
 
   logoff_timer = () ->
 
-    if document.URL.indexOf('session/new') > -1
+    if document.URL.indexOf('session/new') > -1 || document.URL.indexOf('session/sso/inactive') > -1
```
psunix, line 18: create own method for this ?

isLoginPage ?

app/controllers/authentication/user_authenticator.rb
```diff
@@ -52,6 +52,10 @@ def user_logged_in?
     raise NotImplementedError, 'implement in subclass'
   end
 
+  def logged_out_path
```
psunix, line 55: :+1:

psunix: make session_new_path default and override for keyloak ?

Thunderguts13:
Logged in Users who visit /sso/inactive get access denied and redirected. It's "locked" with a pundit policy

app/assets/javascripts/modules/auto_logoff.js.coffee
Thunderguts13: :+1:

app/controllers/authentication/user_authenticator.rb
Thunderguts13: :+1:

PR #254, opened 2020-06-15
system tests

LBruegger:


spec/system/api_users_system_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    click_link 'Api Users'
+
+    within('#api_users_table_body') do
+      #check if table contains api user with updated description
+      expect(page.text).to match('description from api_user')
+
+      #check if possible to lock api_user
+      expect(find('div#active-api-user.toggle-button').click)
+      expect(find('div#active-api-user.toggle-button-selected'))
```
TheWalkingLeek, line 54: are the expected here necessary? If you want to check wether your button press does something you could do it like this right?
```
find('div#active-api-user.toggle-button').click
expect(find('div#active-api-user.toggle-button-selected')).to be_present
```

spec/system/api_users_system_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      # renew user to validate that valid until has values
+      expect(find('img#renew-user.action-icons').click)
+      expect(find("div#valid_until").text) != ''
```
TheWalkingLeek, line 58: Same here. The statement on 58 seems even weirder

PR #270, opened 2020-06-30
Add favourising teams and filtering by favourites, refs: #53708

TheWalkingLeek:


app/controllers/api/teams/favourites_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
+#  Cryptopus and licensed under the Affero General Public License version 3 or later.
+#  See the COPYING file at the top-level directory or at
+#  https://github.com/puzzle/cryptopus.
+module Api
+  module Teams
+    class FavouritesController < ApiController
+
+      def destroy
+        raise 'Team is already not a favourite' if entry.nil?
```
LBruegger, line 12: Perhaps Team isn't favourite already?

app/controllers/api/teams/favourites_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        super
+      end
+
+      private
+
+      def build_entry
+        raise 'Team is already a favourite' if entry.present?
```
LBruegger, line 19: Perhaps same here: Team is favourite already?

app/presenters/teams/filtered_list.rb
```diff
@@ -43,5 +48,9 @@ def filter_by_query
     def filter_by_id
       [Team.find(team_id)]
     end
+
+    def filter_by_favourite
```
LBruegger, line 52: Is this going to be returned as array, if only one favourite selected?

app/serializers/team_serializer.rb
```diff
@@ -22,4 +22,9 @@ class TeamSerializer < ActiveModel::Serializer
   attributes :id, :name, :description, :private
 
   has_many :folders, serializer: FolderMinimalSerializer
+
+  has_many :user_favourite_teams do |serializer|
+    serializer.object.user_favourite_teams.list.where(user: scope)
+    # scope.user_favourite_teams
```
LBruegger, line 28: Comment needed?

config/locales/en.yml
```diff
@@ -315,6 +315,9 @@ en:
       members:
         removed: "%{username} was successfully removed from team"
         added: "%{username} was successfully added to team"
+      favourites:
+        removed: "%{team_name} was successfully removed from the favourites"
+        added: "%{team_name} was successfully added to the favourites"
```
LBruegger, line 320: Perhaps remove the. I think it increases language flow.

config/routes/api.rb
```diff
@@ -43,9 +43,9 @@
       end
 
       resources :folders, except: [:new, :edit, :destroy]
-
       resources :api_users, only: [:create, :destroy, :index], module: 'teams'
       resources :members, except: [:new, :edit], module: 'teams'
+      resource :favourites, only: [:create, :destroy], module: 'teams'
```
LBruegger, line 48: Why resource and not resources?

frontend/app/adapters/user-favourite-team.js
```diff
@@ -0,0 +1,21 @@
+import ApplicationAdapter from "./application";
+
+export default class UserFavouriteTeamAdapter extends ApplicationAdapter {
+  namespace = "api/teams";
+
+  urlForCreateRecord(modelName, snapshot) {
+    return `/${this.namespace}/${snapshot.belongsTo("team", {
+      id: true
```
LBruegger, line 8: Identifier as boolean?

frontend/app/adapters/user-favourite-team.js
```diff
@@ -0,0 +1,21 @@
...
+    })}/${this.pathForType()}`;
+  }
+
+  urlForDeleteRecord(id, modelName, snapshot) {
+    return `/${this.namespace}/${snapshot.belongsTo("team", {
+      id: true
```
LBruegger, line 14: Same here.

spec/controllers/api/team/favourites_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it 'Adds new team to favourites' do
+      login_as(:bob)
+      team2 = teams(:team2)
+
+      expect do
+        post :create, params: { team_id: team2.id }, xhr: true
+      end.to change { bob.favourite_teams.size }.from(1).to(2)
+      relationships = data['relationships']
+      expect(relationships.count).to eq(1)
+      expect(relationships['favourite_team']['data']['id']).to eq(team2.id.to_s)
```
LBruegger, line 20: Check http status.

spec/controllers/api/team/favourites_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+#     it 'doesn\'t remove favourite from user if not team member' do
+#       login_as(:alice)
+#       entry = double()
+#       expect_any_instance_of(Api::Teams::FavouritesController).to receive(:fetch_entry).exactly(:once).and_return(entry)
+
+#       delete :destroy, params: { team_id: teams(:team2) }, xhr: true
+#       expect(response).to have_http_status(403)
+#     end
```
LBruegger, line 66: Delete?

spec/support/controller_helpers.rb
```diff
@@ -33,4 +33,8 @@ def expect_json_object_includes_keys(json, keys)
   def errors
     json['errors']
   end
+
+  def infos
+    json['info']
+  end
```
LBruegger, line 39: Nice 🚀

app/presenters/teams/filtered_list.rb
TheWalkingLeek: It does. Since favourite_teams is a `has_many` relation, it will always return an array

app/serializers/team_serializer.rb
TheWalkingLeek: Well noticed. These two lines do the exact same, I'm actually not completely content with how I have to get the records here. If you got a better idea feel free to help out, was gonna discuss this with @psunix afterwards though. That's why I kept it for the moment

config/routes/api.rb
TheWalkingLeek: Huh, this seems more of a typo. Interesting that it actually works nonetheless. Thanks for the hint

config/locales/en.yml
TheWalkingLeek: We actually don't need these locales anymore, I agree with you though :p

frontend/app/adapters/user-favourite-team.js
TheWalkingLeek: Right. This is kind of confusing, since we only have access to the `snapshot` we have to access the team using `belongsTo()` and because we only want the id we use this Object to specify so.
https://api.emberjs.com/ember-data/3.19/classes/Snapshot/methods/belongsTo?anchor=belongsTo

config/routes/api.rb
TheWalkingLeek: It is not a typo! @Thunderguts13  is simply a genius
https://guides.rubyonrails.org/routing.html#singular-resources

This allows us to use: `DELETE /api/teams/1/favourites`

otherwise we would have needed to use: `DELETE /api/teams/1/favourites/1`

That's really cool :D

app/serializers/team_serializer.rb
psunix: interesting problem :) how to avoid n+1 queries here ?

I would rather create a computed boolean flag if it's favourite or not.

maybe we have to preload all favourite teams first and then set the boolean flag on the entries ?

app/serializers/user_favourite_team_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+# Table name: user_favourite_team
+#
+#  id         :integer          not null, primary key
+#  team_id    :integer          default(0), not null
+#  user_id    :integer          default(0), not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class UserFavouriteTeamSerializer < ActiveModel::Serializer
```
psunix, line 14: is this serializer required ?

config/routes/api.rb
psunix: use api/teams/1/favourite instead of api/teams/1/favourites

it's not possible to have multiple favourites per team ! and in this case it's more an action than resources.

db/migrate/20200623220242_create_user_favourite_team.rb
```diff
@@ -0,0 +1,9 @@
+class CreateUserFavouriteTeam < ActiveRecord::Migration[6.0]
+  def change
+    create_table :user_favourite_teams do |t|
+      t.integer :team_id, default: 0,  null: false
```
psunix, line 4: add_reference :teams, :team, foreign_key: true

frontend/app/adapters/user-favourite-team.js
psunix: instead of creating an adapter here, it might be just ok to use fetch !
a model is not really required, is it?
it's more an action ... favorite this team !

spec/controllers/api/team/favourites_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect do
+        post :create, params: { team_id: team2.id }, xhr: true
+      end.to change { bob.favourite_teams.size }.from(1).to(2)
+      relationships = data['relationships']
+      expect(response).to have_http_status(201)
+      expect(relationships.count).to eq(1)
+      expect(relationships['favourite_team']['data']['id']).to eq(team2.id.to_s)
+    end
+
+    it 'does not create new favourite team without team memberhip' do
```
psunix, line 24: :+1:

spec/unit/models/user_favourite_team_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      user_favourite_team1 = UserFavouriteTeam.new(params1)
+      user_favourite_team2 = UserFavouriteTeam.new(params2)
+      expect(user_favourite_team1).to be_valid
+      expect(user_favourite_team2).to be_valid
+    end
+  end
+
+  context 'new' do
+    it 'remove teammember' do
+      team1_bob = user_favourite_teams(:team1_bob)
```
psunix, line 37: ? what's this for ?
testing if a record can be destroyed ?

spec/controllers/api/teams_controller_spec.rb
```diff
@@ -89,6 +88,31 @@
       expect(folder_relationships_length).to be(3)
     end
 
+    it 'returns favourite teams of bob when favourite is given' do
```
psunix, line 91: it ' returns bob\'s favourite teams' ...

PR #274, opened 2020-07-03
Send current user per env_settings, refs: #54348

TheWalkingLeek:


app/controllers/api/env_settings_controller.rb
```diff
@@ -6,6 +6,6 @@ class Api::EnvSettingsController < ApiController
   before_action :skip_authorization
 
   def index
-    render_json sentry: ENV['SENTRY_DSN_FRONTEND']
+    render_json(sentry: ENV['SENTRY_DSN_FRONTEND'], current_user_id: current_user.id)
```
psunix, line 9: wouldn't it be nice to have other user information too ?

like surname, username, id ?

user: { id: ..., surname: ...., username: ... }

maybe not required currently ... but how about the structure current_user: { id: 42 } ?

PR #278, opened 2020-07-10
Adjust ember implementation and copy assets to frontend, refs: #54643

TheWalkingLeek:


TheWalkingLeek:
What doesn't work yet is the CSRF Token... I'm on it

PR #279, opened 2020-07-10
Feature 54543 rails frontend styling

LBruegger:


frontend/package.json
```diff
@@ -30,9 +30,10 @@
     "ember-auto-import": "^1.5.3",
     "ember-autofocus-modifier": "^1.0.1",
     "ember-bootstrap": "^3.1.4",
+    "ember-bootstrap-power-select": "^2.0.0",
     "ember-changeset": "^3.3.8",
     "ember-changeset-validations": "^3.3.2",
-    "ember-cli": "~3.16.2",
+    "ember-cli": "~3.19.0",
```
TheWalkingLeek, line 36: Huh? Why is this here? Also, there's a bug of powerselect in 3.19. I'd rather we stay on 3.16 since it's also LTS

frontend/app/templates/components/account-form.hbs
```diff
@@ -4,68 +4,76 @@
   </Modal.header>
```
TheWalkingLeek, line 4: Why was this file even touched here? Looks like that Bootstrap stuff... Is this here on purpose?

frontend/app/templates/components/nav-bar.hbs
```diff
@@ -21,7 +21,7 @@
         {{/if}}
```
TheWalkingLeek, line 21: Again, I don't see why this file is even touched. This story is purposefully just for the SSR Frontend.

frontend/app/templates/components/account-form.hbs
LBruegger: Merge conflict.

frontend/app/templates/components/nav-bar.hbs
LBruegger: Again Merge Conflict.

LBruegger:
> image
> 
> I currently only get the translation key when changing a user role. Also, the table header "Admin" doesn't seem right there, shouldn't it be role?

This is not my fault. Still there on master. But I will fix it for Sure.

LBruegger:
> almost there ...
> 
> * if I only select a team, I cannot select any folders from the drop down in the account form. the folder field also has errors then.
> * if only a team is selected, it should pre-select this team, but no errors yet for the folder. the user has to select a folder first.
> 
> code looks good !

I just fixed the Collapsable issues from Folder and Team. This is now not an issue anymore 😉

psunix:
image

since bob is not a ConfAdmin nor a Admin, the Settings dropdown should not be displayed.

PR #280, opened 2020-07-10
Preselect Team and Folder if selected, refs: #54657

LBruegger:


frontend/app/components/account-form.js
```diff
@@ -30,6 +31,8 @@ export default class AccountForm extends BaseFormComponent {
       AccountValidations
     );
 
+    this.presetTeamAndFolder()
```
TheWalkingLeek, line 34: Maybe do the `if (this.isNewRecord)` here. Otherwise, this method would be called `presetTeamAndFolderIfIsNewRecord` right? :P

frontend/app/components/account-form.js
```diff
@@ -43,6 +46,19 @@ export default class AccountForm extends BaseFormComponent {
     });
   }
 
+  presetTeamAndFolder() {
+    let selectedTeam = this.navService.selectedTeam;
+    let selectedFolder = this.navService.selectedFolder;
+
+    if ((selectedTeam !== undefined) && this.isNewRecord){
```
TheWalkingLeek, line 53: Do we need this undefined check? I mean if theres no team selected on the navbar, there is none assigned here. So even if it is undefined, we can just assign it anyways

frontend/app/components/account-form.js
```diff
@@ -81,6 +97,7 @@ export default class AccountForm extends BaseFormComponent {
 
   @action
   setFolder(folder) {
+    this.selectedFolder = folder;
```
TheWalkingLeek, line 100: huh, what's that needed for? I think you can delete the whole instance variable

frontend/app/components/folder-form.js
```diff
@@ -43,6 +46,14 @@ export default class FolderForm extends BaseFormComponent {
     });
   }
 
+  presetTeam() {
+    let selectedTeam = this.navService.selectedTeam;
+
+    if ((selectedTeam !== undefined) && this.isNewRecord){
+      this.selectedTeam = selectedTeam
+    }
```
TheWalkingLeek, line 54: Same stuff here. Maybe you can even just put it inside the constructor directly since you'll just be doing
```
if (this.isNewRecord) this.selectedTeam = this.navService.selectedTeam
```

frontend/app/components/account-form.js
LBruegger: Sure.

frontend/app/components/account-form.js
LBruegger: Yes, we don't need it.

frontend/app/components/account-form.js
LBruegger: You're right.

PR #281, opened 2020-07-13
Feature 54646 footer

TheWalkingLeek:


frontend/app/styles/app.scss
```diff
@@ -75,9 +75,6 @@ $navbar-dark-brand-hover-color: $navbar-dark-brand-color;
   margin-bottom: 10px;
 }
 
-.w-20 {
-  width: 20%;
-}
 .w-20 {
   width: 20%;
 }
```
LBruegger, line 80: ok, nice 😄

frontend/config/environment.js
```diff
@@ -7,6 +7,7 @@ module.exports = function(environment) {
     rootURL: "/",
     locationType: "auto",
     sentryDsn: "",
+    appVersion: undefined,
```
LBruegger, line 10: Undefined for Sure? What about a fallback like V1.0?

TheWalkingLeek: hmm I see your point... With undefined, it just renders nothing. The link just shows
"Version"
instead of
"Version 3.2"
Think that makes more sense than showing a wrong version number :thinking:

app/controllers/api/env_settings_controller.rb
```diff
@@ -1,12 +1,15 @@
...
 
   skip_before_action :validate_user
   before_action :skip_authorization
 
   def index
     render_json(sentry: ENV['SENTRY_DSN_FRONTEND'],
-                current_user: { id: current_user.id })
+                current_user: { id: current_user.id,
+                                preferred_locale: current_user.preferred_locale },
+                version: version_number)
```
psunix, line 13: :love_you_gesture:

psunix:
it still says Version 3.1 ... it's 3.2, isn,t it ? :wink:

psunix:
changelog for 3.2 missing !

psunix:
when I switch to Russian, the whole footer is broken.
is there a way to switch back to English if any translation keys missing ?
that would probably be the nicest and easiest fix

config/routes/api.rb
```diff
@@ -7,6 +7,8 @@
 
     get 'env_settings', to: 'env_settings#index'
 
+    patch 'locale', to: 'locales#update'
```
psunix, line 10: we got session, profile, ... then it should also be called locale, not locales I think

app/controllers/api/locales_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
```
psunix, line 1: locale controller !

PR #282, opened 2020-07-13
Refresh nav bar on changes and fix bugs, refs: #54656

TheWalkingLeek:


frontend/app/components/team-form.js
```diff
@@ -40,8 +41,11 @@ export default class AccountForm extends BaseFormComponent {
   handleSubmitSuccess(savedRecords) {
     this.abort();
     if (this.isNewRecord) {
-      let team = savedRecords[0].id;
-      this.router.transitionTo("teams.show", team);
+      let team = savedRecords[0];
+      let teams = this.navService.availableTeams.toArray();
```
LBruegger, line 45: Not better if available teams in navService would be an Array?

frontend/app/components/team-show.js
```diff
@@ -40,7 +40,18 @@ export default class TeamShowComponent extends Component {
...
+    }).then(() => {
+      this.args.team.favourised = !this.args.team.favourised;
+      if (this.navService.isShowingFavourites) {
+        let teams = this.navService.availableTeams.toArray();
+        if (this.args.team.favourised) {
+          teams.pushObject(this.args.team);
+        } else {
+          teams.removeObject(this.args.team);
+        }
+        this.navService.availableTeams = teams;
```
LBruegger, line 52: We could possibly save some lines here 😉

frontend/app/components/team-form.js
TheWalkingLeek: It is. But it's a MutableArray.
Problem with that is: Ember somehow doesn't like it if you manipulate the query result (Which navService.availableTeams is). So when I push into that, it throws an error. So I need to convert it to a JS Array, then push into that array and then overwrite the navService.availableTeams variable. It's stupid, I don't like it, but it's the only way I figured out how to get around it...

frontend/app/components/team-show.js
TheWalkingLeek: Considering what I wrote above... can we? :sweat_smile:

frontend/app/components/team-form.js
LBruegger: And if you convert the Array when you initialize it? So that we have a JS Array instead of an Ember MutableArray?

TheWalkingLeek: I could just keep it as a JS Array inside the navService... Would also clear confusion since it's sometimes a JS and sometimes a Mutable Array. Will do!

LBruegger: Nice! Thanks 😄

PR #291, opened 2020-07-15
Fix showing newly added file-entry instantly

lucashabersaat:


frontend/app/models/file-entry.js
```diff
@@ -10,12 +10,24 @@ export default class FileEntry extends Model {
...
+    let promise = this.file.upload(url, opts);
+    promise
+      .then( savedRecords => {
+        let data = JSON.parse(savedRecords.body).data
+        this.id = data.id;
+        this.filename = data.attributes.filename;
+      })
+      .catch(() => {
+        this.account = null;
+      });
```
lucashabersaat, line 29: Didn't work in `handleSubmitError` in [file-entry.js](https://github.com/puzzle/cryptopus/blob/55a70697d667a38f3e7029c4b22678d6e5fa63ab/frontend/app/components/file-entry-form.js#L62).

frontend/app/components/account-show.js
```diff
@@ -6,9 +6,14 @@ import { inject as service } from "@ember/service";
...
   @service store;
 
-  fileEntries = this.store.query("file-entry", {
-    accountId: this.args.account.id
-  });
+
+  constructor() {
+    super(...arguments);
+
+    this.fileEntries = this.store.query("file-entry", {
```
TheWalkingLeek, line 13: We don't need to assign this to a variable anymore

frontend/app/models/file-entry.js
TheWalkingLeek: hmm can you not directly set the account on the record inside file-entry-form? Not using the changeset
`this.record.account = null`

lucashabersaat: Thanks for this suggestion, works.

PR #296, opened 2020-07-17
Translate validation errors, refs: #54773

TheWalkingLeek:


frontend/translations/de.yml
```diff
@@ -519,3 +519,19 @@ de:
     team_policy:
       update?:  Du bist nicht Mitglied dieses Teams
       destroy?: Nur Admins dürfen Teams löschen
+
+  validations:
+    name:
+      present: Der Name muss vorhanden sein
+      tooLong: Der Name ist zu lang
+    description:
+      tooLong: Die Beschreibung ist zu lang
```
psunix, line 528: since translation yml is coming from rails/ruby, it uses snake_case from now.
so now there's a mix of different cases.
shouldn't we just use snake_case here?

PR #301, opened 2020-07-20
Feature 54724 rails frontend styling

lucashabersaat:


lucashabersaat:
Has been merged into #302

PR #305, opened 2020-07-21
Fix account validations and add translations

lucashabersaat:


lucashabersaat:
This has been merged into #306

PR #312, opened 2020-07-23
Change Account-Cards to Account-List, refs: #54810

lucashabersaat:


frontend/app/components/account-entry.js
```diff
@@ -0,0 +1,81 @@
+import Component from "@glimmer/component";
```
psunix, line 1: account-row instead of account entry ?

PR #319, opened 2020-08-04
Hide forbidden actions in frontend, refs: #54811

lucashabersaat:


app/serializers/team_serializer.rb
```diff
@@ -19,18 +19,22 @@
...
+  def deletable
+    TeamPolicy.new(user, object).destroy?
   end
 
   private
 
-  def user_id
-    current_user.id
+  def user
+    current_user
```
lucashabersaat, line 37: such that i can mock the current user in the serializer tests

config/locales/ch_be.yml
```diff
@@ -340,6 +340,7 @@ ch_be:
         user_not_logged_in: Benutzer isch nid igloggt
         invalid_request: Invalidi Ahfrag
         record_not_found: Ihtrag isch nid gfunde worde
+        delete_failed: Ihtrag het nid chenne g'löscht wärde
```
TheWalkingLeek, line 343: Uiuiui das söu de aber bärndütsch si ;)

frontend/tests/unit/services/user-service-test.js
```diff
@@ -0,0 +1,12 @@
...
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Service | user-service', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let service = this.owner.lookup('service:user-service');
+    assert.ok(service);
+  });
```
TheWalkingLeek, line 11: Either put some real tests here or delete the whole file.
If the service doesn't have any logic, a unit test seems pretty useless :)

spec/system/teams_page_system_spec.rb
```diff
@@ -7,6 +7,7 @@
 
 require 'rails_helper'
 
+require_relative '../../app/controllers/api/teams/members_controller'
```
TheWalkingLeek, line 10: Why is this needed?

spec/unit/serializers/team_serializer_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
+#  Cryptopus and licensed under the Affero General Public License version 3 or later.
+#  See the COPYING file at the top-level directory or at
+#  https://github.com/puzzle/cryptopus.
+
+require 'rails_helper'
+
+describe TeamSerializer do
```
TheWalkingLeek, line 10: Niiiiiice!!!

spec/system/teams_page_system_spec.rb
lucashabersaat: When testing this spec alone, the members_controller is not defined, however it works when doing all specs. Autoload is a bit messy and this line makes sure, it's loaded. Will insert a comment.

PR #335, opened 2020-09-22
Feature 55319 ose secret sti

TheWalkingLeek:


db/migrate/20200917140436_add_type_to_account.rb
```diff
@@ -0,0 +1,12 @@
+class AddTypeAndEncryptedDataToAccount < ActiveRecord::Migration[6.0]
+  def up
+    add_column :accounts, :type, :string, default: 'Account::Credentials', null: false
+    change_column :accounts, :type, :string, default: nil
```
psunix, line 4: maybe we could just keep the default value? 
still thinking about anything that could go wrong if the type attr has a default value ... nothing coming up so far :joy_cat:

psunix (changes requested):
looks pretty good, lot's of code !

the only no-go is the use of Account specific details inside CryptUtils. That's really bad design like this.

app/models/account/ose_secret.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class Account::OSESecret < Account
+  attr_accessor :data
+
+  def decrypt(team_password)
+    @data = decrypt_data(team_password)
+  end
+
+  def encrypt(team_password)
```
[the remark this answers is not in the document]
TheWalkingLeek: This isn't possible as far as I can see it. Problem is, that you have to assign the encrypted_data field to the Account entry when encrypting, and the ose_secret field when decrypting. You can see how I adjusted it in the review changes :)

app/models/account/encrypted_data.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class Account::EncryptedData
```
psunix, line 3: how about:

cleartext_data, data, iv

#encrypt(encryption_key), #decrypt(encryption_key)

#to_json implement your-self (returns the correct attrs, raises exception if #encrypt was not called -> data is empty)

app/models/account/ose_secret.rb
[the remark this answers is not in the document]
psunix: you're right !

lib/crypt_utils.rb
```diff
@@ -109,6 +109,30 @@ def decrypt_blob(blob, team_password)
       decrypted_blob << cipher.final
       decrypted_blob.force_encoding('UTF-8')
     end
-  end
 
+    def encrypt_base64(data, cipher_key, iv)
```
psunix, line 113: here you called it cipher_key ... further up, it's called encryption_key. 

I prefer encryption_key over cipher_key ... or just key ?

psunix (changes requested):
magnifique !!!

some small things to take care of.

psunix (approved):
:rocket: :partying_face:

PR #336, opened 2020-10-09
Refactor CcliTokenController, refs: #55431

TheWalkingLeek:


app/controllers/api/env_settings_controller.rb
```diff
@@ -14,7 +14,8 @@ def index
                                 preferred_locale: current_user.preferred_locale },
                 last_login_message: last_login_message,
                 version: version_number,
-                csrf_token: form_authenticity_token)
+                csrf_token: form_authenticity_token,
+                base_url: request.base_url)
```
psunix, line 18: is it save to rely on request for getting the base_url? I you thought this through, then it's fine.
there's also the env var RAILS_HOST_NAME which is set on our prod envs and contains the base url

psunix (approved):
:+1: 
one comment to check

PR #394, opened 2021-05-06
Update Ember to 3.24 LTS

luca-bruegger:


frontend/config/ember-cli-update.json
```diff
@@ -0,0 +1,20 @@
+{
```
TheWalkingLeek, line 1: Do we need this? Maybe just remove this file if it isn't necessary

frontend/.eslintrc.js
```diff
@@ -9,19 +9,29 @@ module.exports = {
...
   env: {
     browser: true
   },
   rules: {
     "ember/no-global-jquery": "off",
+    "ember/no-classic-classes": "off",
+    "ember/require-tagless-components": "off",
+    "ember/no-classic-components": "off",
+    "ember/no-component-lifecycle-hooks": "off",
+    "space-before-function-paren": ["off", "always"],
```
TheWalkingLeek, line 26: maybe double-check those rules and look up if they make sense to you

TheWalkingLeek (approved):
Seems like we didn't have prettier installed project wide before...
Better now than never! Try and check through your changed files before creating a PR and creating separate commits for actual update and prettier changes. This was kind of tiresome to review since i wasn't able to separate "real" from prettier changes. It is fine though now

Other than that. Looks good! Was able to run this locally (non docker setup) instantly without any problems. CI proves himself. Maybe check those comments and then merge :)

frontend/.eslintrc.js
luca-bruegger: Those rules are needed for the imports of glimmer and classic classes. I've set them to off, because it's a lot of work to change those import in every file. 
For now though,  I think we are fine if we leave it as it is.

PR #400, opened 2021-05-10
Move profile to ember, refs: #366

### LINKED ISSUE #366: Move profile to ember
The profile is currently still SSR. Needs to be moved to ember. 
Currently, the profile provides different actions using ajax requests and coffeescript, these need to be reworked in ember

All these routes should be removeable afterwards:
```  
get 'profile', to: 'profile#index'
```

TheWalkingLeek:
Closes #366


frontend/app/components/api-users-table-row.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @action
+  toggleApiUser(user) {
+    let httpMethod = user.locked ? "delete" : "post";
+    /* eslint-disable no-undef  */
+    return fetch(`/api/api_users/${user.id}/lock`, {
+      method: httpMethod,
+      headers: {
+        "Content-type": "application/x-www-form-urlencoded; charset=UTF-8",
```
psunix, line 29: we're using this fetch method with the same headers over and over again.
would it make sense to define a global custom fetch method? #make-it-dry

psunix (approved):
nice, looks awesome :partying_face: 

on the delete api confirmation dialog: would it be possible to mention the api username?

`Delete api user alice-bc7bfc ?`

only if easy to get done

PR #420, opened 2021-05-20
Feature 392 users to emberjs

### LINKED ISSUE #392: ADMIN: Move users to emberjs (list)
currently, admin -> users is server side rendered in classic rails app part. the goal of this issue is to move it to api and emberjs. 
image

# Tech-Spec
## Backend
- [x] users#index has to be JSON API

## Frontend
- [ ] UI is loaded in the normal router `{{outlet}}`. Meaning, the Team Navigation is visible and the users are loaded to the right of it 

Just like the red space in the screenshot
- [ ] Reconstruct table in UI

TheWalkingLeek:
Closes #392


app/controllers/api/admin/users_controller.rb
```diff
@@ -19,6 +21,14 @@ def destroy
 
   private
 
+  def fetch_entries
+    if params[:locked] == 'true'
```
psunix, line 25: how about introducing the param converter concern? https://github.com/hitobito/hitobito/blob/master/app/controllers/concerns/param_converters.rb#L12 :man_dancing:

app/controllers/redirected_routes/url_handler.rb
```diff
@@ -14,7 +14,8 @@ class RedirectedRoutes::UrlHandler
     teams_show: /(^\/teams\/\d+)$/,
     folder_show: /(^\/teams\/\d+\/folders\/\d+)$/,
     accounts_show: /(^\/accounts\/\d+)$/,
-    profile: /(^\/profile)$/
+    profile_index: /(^\/profile)$/,
+    admin_users_index: /(^\/admin\/users)$/
```
psunix, line 18: for now this is the way to go.
keep in mind to change this as soon we have migrated most of the UI stuff to ember.
then we probably can work with a more generic approach. (redirect unless /api/*)

frontend/app/adapters/user-human.js
```diff
@@ -1,9 +1,9 @@
 import ApplicationAdapter from "./application";
 
-export default ApplicationAdapter.extend({
+export default class UserHumanAdapter extends ApplicationAdapter {
   pathForType() {
     return "user_humen";
```
psunix, line 5: user_human ?

psunix (approved):
looks veryyy goood!
how about showing a badge or label for every user entry indicating what auth provider it is using?
I think it's ok to merge this into master. so we can detect problems early on integration.

frontend/app/adapters/user-human.js
TheWalkingLeek: Nope, user_humen is correct. This refers to the [type attribute in json api](https://jsonapi.org/format/#document-resource-object-identification) which we always serialize pluralized. In this case, user_humen

PR #421, opened 2021-05-25
Feature 381 profile adjust post locale

LivioLiechti:
Pls check


app/assets/javascripts/profile.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
psunix, line 1: we do not need any more view or js files in backend. so completely remove all profile coffee, scss ... files

app/controllers/api/admin/settings_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
+#  Cryptopus and licensed under the Affero General Public License version 3 or later.
+#  See the COPYING file at the top-level directory or at
+#  https://github.com/puzzle/cryptopus.
+
+class Api::Admin::SettingsController < ApiController
```
psunix, line 8: SettingsController? should this be on this branch?
you're working on profile, not settings ...

app/controllers/api/all_folders_controller.rb
```diff
@@ -8,10 +8,6 @@
 class Api::AllFoldersController < ApiController
   self.custom_model_class = Folder
```
psunix, line 10: what does this change have to do with your feature?

app/controllers/api/profile_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Api::ProfileController < ApiController
+  skip_before_action :validate_user
```
psunix, line 4: why is authorization skipped here?

app/controllers/api/profile_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  before_action :skip_authorization
+
+  self.permitted_attrs = [:preferred_locale]
+
+  def update
```
psunix, line 9: maybe there is an elegant way to accomplish this with dry crud? 
check docs at https://github.com/codez/dry_crud
sync with Lian ... he already red some things about dry crud. 

hint: probably override the def entry method for customizing this.

spec/controllers/api/profile_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+        data: {
+          attributes: {
+            preferred_locale: 'de'
+          }
+        }
+      }
+
+      patch :update, params: preferred_locale_params
+
+      expect(bob.preferred_locale).to eq('de')
```
psunix, line 24: bob.reload.pref .... you need to reload entry bob. 
what was his locale before? make sure it's changed after this call

psunix (changes requested):
most of these changes belong to settings branch lian is working on ?!

this feature is about profile, not settings.

PR #429, opened 2021-06-01
Feature 381 profile adjust post locale

### LINKED ISSUE #381: PROFILE: Adjust POST locale
Currently, we change the locale via a separate endpoint
https://github.com/puzzle/cryptopus/blob/master/config/routes.rb#L34
This should now be merged into the POST /api/profile endpoint and respective controller as described in https://github.com/puzzle/cryptopus/issues/375

 All these routes should be removeable afterwards:
```
  post 'session/locale', to: 'session#changelocale'
```

let locale drop-down use new endpoint

# Tech-Spec

- new endpoint: PATCH /api/profile, api/profile_controller.rb with action #update
- change frontend to point to new endpoint
- completely remove /session/locale endpoint

LivioLiechti:
Closes #381

### DIFF
--- a/config/routes/api.rb
+++ b/config/routes/api.rb
@@ -7,7 +7,7 @@ Rails.application.routes.draw do
 
     get 'env_settings', to: 'env_settings#index'
 
-    patch 'locale', to: 'locale#update'
+    resource :profile, only: [:update]
 
     resources :accounts, except: [:new, :edit] do
       resources :file_entries, only: [:create, :index, :destroy, :show]
--- a/app/controllers/api/profiles_controller.rb
+++ b/app/controllers/api/profiles_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class Api::LocaleController < ApiController
+class Api::ProfilesController < ApiController
   skip_before_action :validate_user
   before_action :skip_authorization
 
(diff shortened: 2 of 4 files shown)


app/controllers/api/profile_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class Api::LocaleController < ApiController
+class Api::ProfileController < ApiController
   skip_before_action :validate_user
   before_action :skip_authorization
```
TheWalkingLeek, line 5: Do we need those (skip_)before_actions?

app/controllers/api/profile_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class Api::LocaleController < ApiController
+class Api::ProfileController < ApiController
```
TheWalkingLeek, line 3: Try to crudify this controller:
- remove update method
- override entry method to just return current_user

That should already do the trick :)

config/routes/api.rb
```diff
@@ -7,7 +7,7 @@
 
     get 'env_settings', to: 'env_settings#index'
 
-    patch 'locale', to: 'locale#update'
+    patch 'profile', to: 'profile#update'
```
TheWalkingLeek, line 10: `resource :profile, only: [:update]` should also do the trick. A bit more consistent with the rest

TheWalkingLeek (changes requested):
Please don't push/commit the test db journal. But looks pretty good already 🐳

PR #430, opened 2021-06-01
Move admin/settings to ember

### LINKED ISSUE #391: ADMIN: Move Settings to emberjs/api
currently, admin -> settings is server side rendered in classic rails app part. the goal of this issue is to move it to api and emberjs. 

# Tech-Spec

## Backend
- [x] Create new controller api/admin/settings_controller.rb, use dry crud, only use index and update actions
- [ ] Create db migration to migrate Settings keys and remove prefix: general_country_source_whitelist -> country_source_whitelist
- [x] Create routes entries, only: index, update
- [x] Create new serializer for setting: app/serializers/setting_serializer.rb (output attrs key, value)
- [x] remove existing SSR settings controller and spec
- [ ] create seeder for both setting entries, make sure seeding is done in development and production env (adjust entrypoint ?)

## Frontend
- [x] Create new Drop-Down Admin in Topbar and add Settings item
- [x] New Route app/routes/admin/settings.js
- [x] New Template app/templates/admin/settings.hbs (inlcudes multiselect for ip_whitelist, country_source_whitelist)
- [ ] New Component app/components/admin/setting_multiselect.js
- [x] Power-select-multiple https://ember-power-select.com/docs/multiple-selection/
- [x] Addon for countries -> Send iso codes to backend
- [ ] No save button, just update settings when hitting enter or leaving input field
  - [ ] In return, each Input in the form needs an `@enter` and `@focus-out` action which save the newly edited data (check if there were changes!)

miosidler:
closes #391


spec/controllers/api/admin/settings_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+          }
+        }, id: country_source_whitelist.id
+      }
+      patch :update, params: settings_params, xhr: true
+      country_source_whitelist.reload
+
+      expect(country_source_whitelist.value).to eq %w[DE US]
+      expect(response).to have_http_status(200)
+    end
+    it 'is not possible for non-admin to update setting' do
```
psunix, line 64: insert blank line between it blocks

spec/controllers/api/admin/settings_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      patch :update, params: settings_params, xhr: true
+      expect(response).to have_http_status(403)
+
+      ip_whitelist.reload
+      expect(ip_whitelist.value).to eq ['0.0.0.0', '192.168.10.0']
+    end
+  end
+  context 'validate setting' do
```
psunix, line 83: blank line ...

frontend/app/components/admin/setting-multiselect-with-create.js
```diff
@@ -0,0 +1,75 @@
...
+      });
+    } else {
+      this.selectedOptions = this.args.selectedOptions;
+    }
+  }
+
+  @action
+  updateSetting(selected) {
+    this.selectedOptions = selected;
+    this.settingRecord.value = selected.map((option) => {
```
psunix, line 39: image

psunix: always check browser console for errors

psunix (changes requested):
wow, lot's of changes and it looks pretty nice already!
some things to do:
- when I add an value (ip or country), it is not saved. when re-visiting the page, all options are gone
- remove app/views/admin/settings
- make sure no js errors in browser console
- hide team navigation when showing settings (see users admin branch), add back button
- after adding an option in the settings field, the option drop down is still open. can't we close this after adding an IP or country option?
image

app/controllers/authentication/source_ip_checker.rb
```diff
@@ -64,7 +64,7 @@ def country_authorized?
   end
 
   def whitelisted_country_codes
-    @whitelisted_country_codes ||= Setting.value(:general, :country_source_whitelist)
+    @whitelisted_country_codes ||= Setting.value('', :country_source_whitelist)
```
psunix, line 67: shouldn't we adjust the value method? getting rid of the prefix?

psunix (approved):
@TheWalkingLeek I added a check for making sure, geoip db is installed. could you do a review on that too?
please check the question about the locale
otherwise ... looks great !  :dancers:

frontend/app/components/admin/setting-multiselect-with-create.js
psunix: those errors popped up because the settings were not available in db! had to reseed

frontend/translations/en.yml
```diff
@@ -110,7 +110,19 @@ en:
         title: Edit Settings
         save: Save changes
         ldap: Test LDAP connection
+        whitelisted_countries: Whitelisted Countries
```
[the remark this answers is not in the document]
TheWalkingLeek: yup... currently we're still translating flash messages in the backend

PR #434, opened 2021-06-03
Sort teams ignoring uppercase and umlauts

carlobeltrame:
In my cryptopus account, the `hitobito` team is listed after the `Quartieridee` and `UZH` teams. This PR changes the sorting algorithm to disregard upper-/lowercase. It also uses `localeCompare` such that umlauts (ä, ö, ü, é, ...) are sorted correctly based on the current browser locale. This is a built-in feature of ECMAScript and supported by all browsers: https://caniuse.com/localecompare

(I haven't manually tested my implementation, please do test it during the review)


TheWalkingLeek:
Works! Thanks a lot :)

Issue #380, opened 2021-04-26
RECRYPT_REQUESTS: Remove recryptrequests


TheWalkingLeek:
Completely remove the recryptrequests routes including back- and frontend logic

Following routes are affected:
```
    resources :recryptrequests, only: [:index, :destroy] do
      collection do
        post 'resetpassword'
      end
    end
```

Steps:
- [x] Look what is entailed in recryptrequests
- [x] Remove everything that includes recryptrequests
- [x] Get tests running again

PR #437, opened 2021-06-04
Remove recrypt requests, refs: #380

luca-bruegger:


Gemfile.lock
```diff
@@ -473,4 +473,4 @@ DEPENDENCIES
   webpacker
 
 BUNDLED WITH
-   2.2.17
+   2.2.18
```
psunix, line 476: is this change related to this feature?
if updating here, probably have to update in docker image too

app/controllers/recrypt/ldap_controller.rb
```diff
@@ -36,21 +36,13 @@ def authorize_action(action)
   end
 
   def create_recrypt_request
```
psunix, line 38: remove this method completely including "forgot password' functionality

config/locales/ru.yml
```diff
@@ -85,28 +85,6 @@ ru:
...
         save: "Сохранить настройки"
-    recryptrequests:
-      index:
-        title: "Запросы на перешифрование"
-        user: "Пользователь"
-        admin_required: "Необходимы права администратора"
-        recrypt: "Перешифровать"
-      request:
-        request_sent: "Запрос на перешифрование ваших групповых паролей был отправлен. Подождите, пока администратор не перешифрует ваши пароли."
-      uncrypterror:
```
psunix, line 96: I think this part is still required, isn't it?
so if the user has changed his ldap password, cryptopus wants to recrypt the private key with the new password.

db/migrate/20210604065657_drop_recryptrequest.rb
```diff
@@ -0,0 +1,6 @@
+class DropRecryptrequest < ActiveRecord::Migration[6.1]
```
psunix, line 1: DropRecryptRequests

db/migrate/20210604065657_drop_recryptrequest.rb
```diff
@@ -0,0 +1,6 @@
...
+  def change
+    drop_table :recryptrequests
+    remove_column :user, :recryptrequests
```
psunix, line 4: there is no column named recryptrequests, is there ?

spec/integration/user_provides_new_ldap_pw_spec.rb
```diff
@@ -123,39 +112,6 @@
       expect(flash[:error]).to match(/Your OLD password was wrong/)
     end
 
-    it 'provides new ldap password and entered wrong new password' do
```
psunix, line 126: this functionality should still be available
just the forgot password functionality is being dropped

psunix (changes requested):
you're on the right way :+1: 

keep in mind ... it should still be possible for a ldap user to change his password.

PR #438, opened 2021-06-09
Feature 428 admin users table styling improvements

LianVin:


frontend/app/templates/components/admin/user/table-row.hbs
```diff
@@ -6,12 +6,12 @@
   <td>{{@user.label}}</td>
   <td>{{if @user.lastLoginAt (moment-format @user.lastLoginAt "DD.MM.YYYY hh:mm") ""}}</td>
   <td>{{@user.lastLoginFrom}}</td>
-  <td>{{@user.providerUid}}</td>
+  <td>{{@user.auth}}{{#if @user.providerUid}}:{{/if}} {{@user.providerUid}}</td>
```
TheWalkingLeek, line 9: Looks kinda strange without the line breaks around the `#if` no?

frontend/tests/integration/helpers/format-role-test.js
```diff
@@ -0,0 +1,17 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | format-role', function(hooks) {
```
TheWalkingLeek, line 6: What? Why is this here? There isn't even a format-role helper

frontend/app/templates/components/admin/user/table.hbs
```diff
@@ -1,12 +1,12 @@
...
-    <div class="col-auto">
-      <button class="btn btn-primary" {{on "click" this.toggleUserNew}}>{{t "new"}}</button>
-    </div>
-  </div>
+<div class="row justify-content-end">
+      {{!-- <button class="btn btn-primary" {{on "click" this.toggleUserNew}}>{{t "new"}}</button> --}}
+        <span class="mx-2" role="button" {{on "click" this.toggleUserNew}}>
+          <img class="icon-big-button" src="/assets/images/plus.svg" alt="new">
+          <BsTooltip @title={{t "tooltips.add_folder"}} @delayShow="1000" />
+        </span>
```
TheWalkingLeek, line 9: This looks somewhat worse like this no? The tooltip is not correct and maybe you could give it a margin from the right.
Right now it just looks a bit randomly placed without real context (at least to me)

frontend/app/templates/admin/users.hbs
```diff
@@ -1,5 +1,12 @@
 <div class="pt-3">
-  <h1>{{t "admin.users.index.title"}}</h1> 
+  <div class="justify-content-end">
+    <LinkTo @route="index">
+      <button type="button" class="close" aria-label="Close">
+        <span aria-hidden="true">&times;</span>
+      </button>
+    </LinkTo>
```
TheWalkingLeek, line 7: Maybe style the back-button like on the account-show page. https://github.com/puzzle/cryptopus/blob/5e89e8b0fd472a01764d980a1825f1ee296a4027/frontend/app/templates/components/account/show.hbs#L10

TheWalkingLeek (changes requested):
Already looks waaay better. Good work!
Still got some inputs :)

luca-bruegger:
So far finished, but there are still the one and other things missing:
- [ ] Test the admin/users component within admin/users-test.js
- [ ] Fix admin/table test. Entailed in admin/user/table-test.js

Issue #379, opened 2021-04-26
CHANGELOG: Move changelog to ember


TheWalkingLeek:
The Changelog is currently still in the SSR part of cryptopus and has to be moved to ember.

All these routes should be removeable afterwards:
```
  get 'changelog', to: 'changelog#index'
```

https://github.com/puzzle/cryptopus/blob/master/CHANGELOG.md

# Tech-Spec
- Read CHANGELOG.md an render as HTML
- add CHANGELOG.md during frontend build to assets
- make sure changelog can be rendered in development
- render changelog entries inside modal dialog

# Realization
- [x] Remove backend Changelog things.
- [x] Create Frontend Modal
- [x] Create Symlink and add alter scripts to copy & delete Changelog.md in public/assets/ (before build) files under bin
- [x] Find & Evaluate md render Addon
- [x] Render md in Modal with Addon

PR #442, opened 2021-06-11
Ember changelog, refs: #379

luca-bruegger:


bin/prepare-frontend.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+cp CHANGELOG.md frontend/markdown/
```
TheWalkingLeek, line 2: Do we need to remove the symlink before copying or does this override?

frontend/app/components/changelog.js
```diff
@@ -0,0 +1,5 @@
+import Component from '@ember/component';
```
TheWalkingLeek, line 1:
```suggestion
import Component from '@glimmer/component';
```
ember octane prefers glimmer :)

frontend/app/templates/components/footer.hbs
```diff
@@ -1,6 +1,12 @@
+{{#if this.isChangelogOpen}}
+  <Changelog @onAbort={{this.toggleChangelogModal}} @title="Changelog"/>
```
TheWalkingLeek, line 2: We don't need to translate the title here right? Probably fine

frontend/package.json
```diff
@@ -61,6 +62,7 @@
     "ember-intl": "^4.4.0",
     "ember-islands": "^1.5.1",
     "ember-load-initializers": "^2.1.2",
+    "ember-markdown-html": "^0.0.1",
```
TheWalkingLeek, line 65: We don't need both addons right? We're using the showdown addon right? Looks better

public/frontend-index.html
```diff
@@ -1 +0,0 @@
-../frontend/dist/index.html
```
TheWalkingLeek, line 1: Please run frontend:clean before pushing

TheWalkingLeek (changes requested):
Gooooood stuff, small small notes

frontend/app/templates/components/footer.hbs
luca-bruegger: It is the same in every language.

bin/prepare-frontend.sh
luca-bruegger: cp, by default, overwrites.

PR #451, opened 2021-06-21
Feature users styling actions refs #450

### LINKED ISSUE #450: USERS: styling, actions
in the current 4.0 RC, there's some things to check:

1. Action Buttons

image

no action buttons shown

2. Add Userr Button

Add user button should not be shown if ldap or oidc as auth provider

3. Title

Instead Cryptopus users, just call it 'Users'

4. Role Update
make sure:

* admins can change all the user roles
* conf admin can only change user to conf admin and versa
* nobody can change his own role
this should be already checked in backend ... make sure frontend has the same behaviour

miosidler:
closes #450

### DIFF
--- a/app/policies/user/human_policy.rb
+++ b/app/policies/user/human_policy.rb
@@ -28,10 +28,6 @@ class User::HumanPolicy < ApplicationPolicy
     end
   end
 
-  def new?
-    AuthConfig.db_enabled? && admin_or_conf_admin?
-  end
-
   def create?
     AuthConfig.db_enabled? && admin_or_conf_admin?
   end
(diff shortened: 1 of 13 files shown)


app/policies/user/human_policy.rb
```diff
@@ -19,7 +19,7 @@ def update?
   def edit?
     return false if own_user?
 
-    unless user.ldap?
+    if user.auth_db?
```
psunix, line 22: :+1:

frontend/app/components/admin/user/table-row.js
```diff
@@ -23,8 +21,31 @@ export default class AdminUserTableRowComponent extends Component {
 
   constructor() {
     super(...arguments);
+  }
+
+  get isRoleEditingDisabled() {
+    if (
+      ENV.currentUserId == this.args.user.id ||
```
psunix, line 28: extract to method isOwnUserOrRoot ?

psunix (approved):

Issue #447, opened 2021-06-17
Puzzle-Shell Improvements


luca-bruegger:
Some small tweaks are needed for the current implementation of the Puzzle Shell.

# Tech Spec
- [x] Reorder Topbar Actions and new Icons
- [x] Change Colors
- [x] Move Dropdown from Administration to User
- [x] Profile as Dropdown Entry aswell as Logout

## Approach
- [x] Select needed Colors
- [x] Replace them according to the Concept
- [x] Delete Logout as spare Button
- [x] Add Logout into Dropdown
- [x] Move Topbar Actions around
- [x] Fix Dropdown

PR #466, opened 2021-06-24
Fix dropdowns and links, refs: #447

luca-bruegger:


TheWalkingLeek (approved):
Looks good!
Is the "Add" dropdown supposed to still be a BsDropdown or also be converted?
https://github.com/puzzle/cryptopus/blob/master/frontend/app/templates/components/nav-bar.hbs#L30

PR #470, opened 2021-06-25
Feature move password update refs #375

### LINKED ISSUE #375: PROFILE: Move update password to profile
Move the db user password updating to the profile. 
Use a new route called POST /api/profile to send the old and new password. The logic should be the same as in https://github.com/puzzle/cryptopus/blob/2e153bb8c82781ec4e6f59e85355adfa3311fd6a/app/controllers/session_controller.rb#L59
Make sure to remove this method and the view files afterwards!
All these routes should be removable afterwards:
```
  get 'session/show_update_password', to: 'session#show_update_password'
  post 'session/update_password', to: 'session#update_password'
```

for any db user, it should now be possible to change his password on his profile page. 
create a short concept/mockup to see how we place the required fields

# Tech-Spec
- endpoint: PATCH /api/profile
- move specs https://github.com/puzzle/cryptopus/blob/master/spec/controllers/session_controller_spec.rb#L8 to frontend/feature specs
- make sure updating password is only possible for db users !
- error handling: password repetition does not match, old password does not match

miosidler:
closes #375


app/controllers/api/profile/password_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+class Api::Profile::PasswordController < ApiController
+  self.permitted_attrs = [:old_password, :new_password1, :new_password2]
+
+  before_action :authorize_action
+
+  def update
+    if password_params_valid?
+      current_user.update_password(old_password,
+                                   new_password)
+      add_info('flashes.session.new_password_set')
```
psunix, line 12:
```suggestion
      add_info('flashes.profile.new_password_set')
```
move locale keys

app/controllers/api/profile/password_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      ok_status = 200
+    end
+    render_json(nil, ok_status || 400)
+  end
+
+  def password_params_valid?
+    return if current_user.is_a?(User::Api)
```
psunix, line 19: maybe integrate this check into policy? so this would return a 403 in case an api user tries to access this endpoint ...
add a test case for this

app/controllers/session_controller.rb
```diff
@@ -56,17 +56,6 @@ def show_update_password
     render :show_update_password
```
psunix, line 56: also remove show_update_password

app/policies/profile/password_policy.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Profile::PasswordPolicy < ApplicationPolicy
```
psunix, line 3: instead of creating a policy for Profile::PasswordPolicy here, you could integrate this into User policy since this action is user related. action would be update_password
maybe check resetpassword action in User Human Policy

config/routes/api.rb
```diff
@@ -5,7 +5,11 @@
 
     get 'env_settings', to: 'env_settings#index'
 
-    resource :profile, only: [:update]
+    resource :profile, only: [:update] do
+      collection do
+        patch :password, to: 'profile/password#update'
```
psunix, line 10: :+1:

spec/controllers/api/profile/password_controller_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+                        new_password1: 'test',
+                        new_password2: 'test' }
+        }
+      }
+      patch :update, params: password_params
+
+      expect(response).to have_http_status(403)
+    end
+
+    it 'returns unauthorized if oidc user tries to update password' do
```
psunix, line 69: :+1:

psunix (changes requested):
almost there .. some things to check :woman_dancing:

PR #484, opened 2021-07-01
Fix starting the Docker setup for the first time

carlobeltrame:
When I started the Docker setup for the first time using `docker-compose up`, I ran into a problem that `/seed` and `/myapp/tmp/*` couldn't be written to by Rails. This prevents the Docker setup from starting correctly.

The reason was that the `tmp` directory is mounted into the containers in docker-compose.yml, but this `tmp` directory is not present in the git repository. If docker-compose has to mount a nonexisting directory into a container, it will create that directory on the host, with root as owner. And since rails doesn't run as root, it later has no permissions to write to these newly created directories.

The solution is simply to add the `tmp` directory including any subdirectories that are mentioned in `docker-compose.yml` to the git repository. This way, they are already present and belong to the user (rather than root) when starting the containers for the first time, and docker-compose will not create them as root.


TheWalkingLeek:
Interesting, didn't know that. Thanks for fixing this!

PR #499, opened 2021-12-21
Fix account creation bug on dashboard

luca-bruegger:


luca-bruegger:
Double checked the changes in-app. Using the store by itself, makes sure the records are cached. Nonetheless theres a second request made, as soon you leave the dashboard and enter the Main-View. 

I tried to find a solution for maybe doing that another way, but I stumbled across the following statement within the ember docs: 

```Text
One downside to returning a cached record is you may find the state of the data has changed
since it was first loaded into the store's identity map. In order to prevent this stale data from
being a problem for long, Ember Data will automatically make a request in the background each
time a cached record is returned from the store. 
```
-> https://guides.emberjs.com/release/models/#toc_caching

The reason the second request is made, is simply to keep the identity map up to date.

PR #505, opened 2022-01-04
Create EncryptedData

luca-bruegger:


spec/migrations/encryptables_data_structure_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  let(:migration) { described_class.new }
+
+  let(:folder1) {folders(:folder1)}
+
+  let(:account1) {accounts(:account1)}
+  let(:account2) {accounts(:account2)}
+  let!(:account3) do
+    Account::Credentials.create!(name: 'spacex', folder: folder1, encrypted_data: {
+      password: {data: '', iv: nil},
```
mtnstar, line 17: we do not need an account with blank values inside encrypted data.
we need a LegacyAccount::Credentials with blank/nil values in :password, :username, since this is what exists today and we wanna make sure this gets migrated correctly

spec/migrations/encryptables_data_structure_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    ActiveRecord::Migration.verbose = verbose
+  end
+
+  around do |test|
+    silent { test.run }
+  end
+
+  context 'up' do
+    before { migration.down }
```
mtnstar, line 36: create LegacyAccount::Credentials in here with blank values for :username, :password

spec/migrations/encryptables_data_structure_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    it 'migrates blob credentials to base64' do
```
mtnstar, line 38:
```suggestion
    it 'migrates blob credentials to EncryptedData with base64 encoding' do
```

spec/migrations/encryptables_data_structure_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+      account3.decrypt(team1_password)
+
+      expect(account3.cleartext_username).to eq(nil)
+      expect(account3.cleartext_password).to eq(nil)
+
+      expect do
+        account1.username
+        account1.password
+      end.to raise_error(ActiveModel::MissingAttributeError)
```
mtnstar, line 77: nice that you're checking if those attributes have been removed!
I think I would rather do something like this:
`Account.attribute_names.exclude?(:username)`

mind, the way you checked it does only confirm that username does not exist anymore. Line 76 is never executed.

spec/migrations/encryptables_data_structure_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  context 'down' do
+    after { migration.up }
+
+    it 'reverts to previous schema' do
+      migration.down
+
+      # account 1
+
+      account1.reload
+      legacy_account = UseEncryptedDataForAccountCredentials::
```
mtnstar, line 91: doesn't this work without UseEncryptedDataForAccountCredentials ?

mtnstar (changes requested):
nice! you are almost there. Some awesome and very clean code.
Just a few things to take a look at. go go go :rocket:

db/migrate/20220104140658_use_encrypted_data_for_account_credentials.rb
```diff
@@ -0,0 +1,67 @@
+class UseEncryptedDataForAccountCredentials < ActiveRecord::Migration[6.1]
```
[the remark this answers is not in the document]
luca-bruegger: I didn't put it in there as it isn't present in other migration files as well.

spec/migrations/encryptables_data_structure_spec.rb
luca-bruegger: No, this raises an error: 
``` ruby    
NameError:
       uninitialized constant LegacyAccount
```

PR #506, opened 2022-01-10
OSESecret data structure

luca-bruegger:


spec/unit/models/encrypted_data_spec.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
+#  Copyright (c) 2008-2022, Puzzle ITC GmbH. This file is part of
```
mtnstar, line 3: no license headers @cryptopus

spec/unit/models/account/ose_secret_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+
+  private
+
+  def decoded_secret
+    Base64.strict_decode64(FixturesHelper.read_account_file('example_secret.secret'))
+  end
+
+  def encrypted_data_json(legacy = false, legacy_hash_key = :ose_secret)
```
mtnstar, line 73: because it makes the code more readable, create own method for legacy data.
#encrypted_data_legacy_json

spec/unit/models/account/ose_secret_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    it 'decrypts encrypted_data ose secret within Hash with old key' do
+      ose_secret_account.encrypted_data = ::EncryptedData.new(encrypted_data_json(true))
+
+      ose_secret_account.decrypt(team1_password)
+
+      expect(ose_secret_account.ose_secret).to eq(decoded_secret)
+    end
+
+    # TODO: correct behaviour for legacy hash?
+    it 'returns nil when key other than ose_secret' do
```
mtnstar, line 57: this case is already tested by EncryptedData's spec, isn't it?

spec/unit/models/account/ose_secret_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    it 'decrypts encrypted_data ose secret within Hash' do
+      ose_secret_account.encrypted_data = ::EncryptedData.new(encrypted_data_json)
+
+      ose_secret_account.decrypt(team1_password)
+
+      expect(ose_secret_account.ose_secret).to eq(decoded_secret)
+    end
+
+    it 'decrypts encrypted_data ose secret within Hash with old key' do
```
mtnstar, line 48:
```suggestion
    it 'decrypts openshift secret with legacy EncryptedData' do
```

spec/unit/models/account/ose_secret_spec.rb
mtnstar: couldn't we call the key legacy_ose_secret if there's still the old data structure?
so for newly created secrets, we call it ose_secret. 
then it's pretty easy to differ between current and legacy encrypted data, isn't it?

spec/migrations/use_encrypted_data_for_account_ose_secret_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    it 'reverts to previous schema' do
+      ose_secret_account = Account::OSESecret.create!(name: 'secret1',
+                                                      folder_id: folder1.id,
+                                                      ose_secret: decoded_secret)
+
+      ose_secret_account.encrypt(team1_password)
+      ose_secret_account.save
+
+      pre_mig_iv = Base64.strict_encode64(ose_secret_account.encrypted_data[:ose_secret][:iv])
+
```
mtnstar, line 61: shouldn't we also test the migration back of an ose secret with legacy datastructure ?

mtnstar (changes requested):
looks good and you took care of many things. some things to check though. 

couldn't we already introduce a Account::DeprecatedOSESecret class?

PR #543, opened 2022-02-10
Refactor CryptUtils

luca-bruegger:


app/controllers/session_controller.rb
```diff
@@ -78,7 +78,7 @@ def create_session(password)
       user = user_authenticator.user
       set_session_attributes(user, password)
       user_authenticator.update_user_info(request.remote_ip)
-      CryptUtils.validate_keypair(session[:private_key], user.public_key)
+      Asymmetric.validate_keypair(session[:private_key], user.public_key)
```
mtnstar, line 81:
```suggestion
      Crypto::Asymmetric.validate_keypair(session[:private_key], user.public_key)
```

app/crypto/asymmetric.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class Asymmetric
```
mtnstar, line 3:
```suggestion
class Crypto::RSA
```
maybe RSA is better in that case. it's shorter.
only Asymmetric is not an option since it does not include information about encryption. 
e.g. there's asymmetric downloads

app/crypto/hashing.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'digest/sha1'
+
+class Hashing
+  def self.hash(data)
```
mtnstar, line 6:
```suggestion
  def self.salted_hash(data)
```

app/crypto/symmetric/AES256.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'openssl'
+require 'digest/sha1'
+
+class Symmetric::AES256 < Symmetric
+  @@CYPHER = 'aes-256-cbc'
```
mtnstar, line 7: maybe define those as class_attributes ?

spec/crypto/asymmetric_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+require 'spec_helper'
+
+describe Asymmetric do
+  let(:keypair) { Asymmetric.generate_new_keypair }
+  let(:private_key) { keypair.to_s }
+  let(:public_key) { keypair.public_key.to_s }
+  let(:team_password) { Symmetric::AES256.random_key }
+
+  it 'should encrypt and decrypt team password' do
```
mtnstar, line 11:
```suggestion
  it 'encrypts and decrypts data' do
``

File: spec/crypto/asymmetric_spec.rb [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+    encrypted_password = Asymmetric.encrypt(team_password, public_key)
```
mtnstar, line 12: cleartext_data ?

mtnstar: now we're abstracting the whole crypto stuff. 
so it doesn't matter what data we're encrypting or decrypting ...
so the whole Crypto Class doesn't have to know about any team passwords

spec/crypto/hashing_spec.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Hashing do
+  let(:password) { 'foo password' }
+
+  it 'should create SHA512 password hash' do
```
mtnstar, line 8:
```suggestion
  it 'creates SHA512 password hash' do
```
why should?

spec/crypto/hashing_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+    hash = Hashing.hash(password)
+    expect(hash).to match(/^sha512\$.*\$.*/)
```
mtnstar, line 10: I find it quite important to check if the generated hash is working correctly.

how about moving https://github.com/puzzle/cryptopus/blob/master/app/models/user.rb#L66 also to Crypto::Hashing ?
#matches?(salted_hash, password)

mtnstar (changes requested):
good good ... go go :smile_cat:

spec/unit/utils/crypto/symmetric/aes256iv_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  let(:keypair) { Crypto::RSA.generate_new_keypair }
+  let(:private_key) { keypair.to_s }
+  let(:public_key) { keypair.public_key.to_s }
+  let(:team_password) { described_class.random_key }
+
+  it 'should encrypt and decrypt data' do
+    key = team_password
+    data = Base64.strict_decode64('test')
+    encrypted_values = described_class.encrypt(data, key)
+    data_encrypted = encrypted_values.first
```
mtnstar, line 17: `data_encrypted, iv = encrypted_values`

mtnstar: encrypted_data better than data_encrypted ?

mtnstar (changes requested):
looks pretty good, almost there!
some things to check ... go go go

PR #546, opened 2022-02-14
Use LinkTo helper for account links

luca-bruegger:


mtnstar:
link is working, but it completely broke account username, password ... no values are shown, neither at account#list nor account#show

image

mtnstar (changes requested):
changes should not break application ...

PR #574, opened 2022-03-17
512 tech upgrade to ruby 3.0, rails 7.0

Drinion:
Running `rake` without `bundle exec` leads to the issue, that the activated rake version is older than the one specified by the `Gemfile`. This is an interesting issue, because Docker fetches the `ruby:3.0` image, which installs `rake 13.0.3`. At the same time, there must be a gem in the `Gemfile` that implicitly installs `rake 13.0.6`. A similar error (different rake versions) seems to be present on the master branch. The issue can be avoided by executing `bundle exec rake` instead of `rake`.


luca-bruegger (changes requested):
Some really good things changed! :rocket: I like the style of refactoring, really makes the code more obvious. 

Maybe look into those things:
- I know classes in Ruby need to be named in _CamelCase_, but doesn't it make sense to name the encryption classes in _UPPERCASE_? As the algorithms `AES256` and `AES256IV` are named uppercase aswell?
- There are some controllers where the `status_code` has been changed. What is the reason for using e.g. the word `:no_content` over the effective status code `204`?
- Within `schema.rb` the keyword `precision: nil` was added automatically, I assume. Could there be any problems regarding the current DB-Schema, by using the precision attr?

Lastly, let @Robin481 check those changes. Just to be sure, that those changes work as they should.

Drinion:
Hello @luca-bruegger ! Thanks for your suggestions.

- When I updated ruby and rails and tried to run the specs, rspec threw an error saying, that `AES256` (and `AES256IV`) don't exist on the module `Crypto::Symmetric`. The same thing happened with `OSE` and rspec asked, if I meant `Aes256` (`Aes256iv`). So I assumed, that these method names must have been changed in some gem outside of Cryptopus.
- These changes were suggested by Rubocop and it throws a formatting error if `204` is used instead of `:no_content`. Should we disable this rule?
- I don't think so because according to the documentation, this simply defines the precision of the given number. 

> ...precision Specifies the precision for decimal/numeric/datetime/time columns

luca-bruegger:
> Hello @luca-bruegger ! Thanks for your suggestions.
> 
> * When I updated ruby and rails and tried to run the specs, rspec threw an error saying, that `AES256` (and `AES256IV`) don't exist on the module `Crypto::Symmetric`. The same thing happened with `OSE` and rspec asked, if I meant `Ose`. So I assumed, that these method names must have been changed in some gem outside of Cryptopus.
> * These changes were suggested by Rubocop and it throws an error if `204` is used instead of `:no_content`. Should we disable this rule?
> * I don't think so because according to the documentation, this simply defines the precision of the given number column and it is only used with `datetimes` (and it's `nil` anyways).
> 
> > ...precision Specifies the precision for decimal/numeric/datetime/time columns

Hey @Drinion thanks for your detailed response

- Regarding that error, you need to use `require_dependency` within the files where the error appears.
- As we have used status codes until now, I would recommend leaving them as numbers. Despite that, I would change it to the status word as it is a valid Rubocop rule and that matters in my opinion. What do you think?
- Ok, thats fine for me.

Robin481:
Heya @Drinion 
I'll have a look at this as soon as the above topics from/with Luca are resolved.
However, a little side note:
Don't feel pressured to now hurriedly finish all of this by tomorrow, it is completely reasonable to not finish this, plan some more time next sprint and finish it without time pressure (as of course my review will also take some time to do).

Drinion:
Hey @Robin481 thanks! The MR should be ready now. Hi @luca-bruegger ! I have changed the class name from snake case back to block case. Using `require_relative` forced me to change the filenames from `aes256.rb` to `AES256.rb` (same for `aes256iv.rb`). Rubocop doesn't like that so I disabled the filename rule (which wasn't declared in the `.rubocop.yml` file). It's unconventional to create `.rb` files with capital letters but makes sense in this case I think. 
About the `:no_content` vs. `:204` issue: I really like the former better, because it's more declarative and basically says the same thing as `:204`.

Drinion:
Hi @luca-bruegger ! Using `require_relative` forces us to rename the file `aes256.rb` to `AES256.rb` which is not recommended. Rubocop also complains about the class name not being in snake case. I agree, it would be more consistent to name the class in uppercase but I think doing so leads to too many convention violations. What do you think?

Robin481:
Heya
I rebased to master as this commit: https://github.com/puzzle/cryptopus/commit/d5d14077f5fca2b5e1e9afc4f7a4e49dab75abdd was needed to fix the development docker env :)

Drinion:
@Robin481 Thanks!

app/controllers/authentication/user_authenticator/oidc.rb
```diff
@@ -42,7 +42,6 @@ def user_passphrase
 
   def find_or_create_user
     user = User.find_by(username: oidc_username.strip)
-    return if user&.is_a?(User::Api)
```
Robin481, line 45: Why was this deleted?

Robin481 (changes requested):
Added a new commit to fix the `last_login_message` translation.
Besides a little question most of it looks good to me.
@mtnstar will also have a look at this as he has the most domain knowledge.
Cheers :beers:

app/controllers/authentication/user_authenticator/oidc.rb
Drinion: Because rubocop threw a ` W: [Correctable] Redundant safe navigation detected.` error.

Robin481: Haha well deleting the Code that Rubocop is complaining about solves about 100% of its complaints :joy: 
Just so you understand:
The safe navigation operator used in this case made sure that when the `user` is nil instead of breaking the application with something along the lines of `is_a? is not a function for "nil"` it just returns false.
It seems that newer versions of Ruby/Rails handle `is_a?` functions more gracefully than before, hence Rubocop started complaining. (I tested this locally in a console by just doing: `nil.is_a?(User)` and this seems to work just fine, henceforth making the safe navigator redundant as correctly prognosed by Rubocop).

I see you already corrected this and correctly removed the safe navigation operator, good job :ghost:

Drinion: Deleting the code is always the most effective solution 😆

yarn.lock
```diff
@@ -2,3 +2,7 @@
 # yarn lockfile v1
 
 
+ember@^1.0.3:
```
mtnstar, line 5: why is there a change in the yarn.lock?

it's not related to rails/ruby upgrade, is it ?

app/models/encryptable/ose_secret.rb
```diff
@@ -2,7 +2,7 @@
 
 require_relative '../../utils/crypto/symmetric/aes256iv'
 
-class Encryptable::OSESecret < Encryptable
+class Encryptable::OseSecret < Encryptable
```
mtnstar, line 5: if we rename this, we need to create a db migration to change given values in encryptables#type column.

could u please create a migration for this?

mtnstar (changes requested):
thx for the great work, looks pretty good and is running well locally. :partying_face: 

we need to create a db migration to adjust given encryptable type values. (Encryptable::OseSecret)

app/models/encryptable/ose_secret.rb
mtnstar: similar to: https://github.com/puzzle/cryptopus/blob/master/db/migrate/20220113085536_rename_accounts_to_encryptables.rb#L13

mtnstar (approved):
awesome :partying_face:

PR #575, opened 2022-03-18
Feature/541 provide sorting in admin/users view

RamonaChristen:
admin/user table can now be sorted by each attribute by clicking on the small arrow next to it in the table header. Attributes are initially sorted in ascending order, a second click sorts in descending order - except for the "last login at" attribute which is initially sorted in descending order to display most recent logins at the top.

### DIFF
--- a/frontend/app/components/admin/user/table.js
+++ b/frontend/app/components/admin/user/table.js
@@ -9,7 +9,7 @@ export default class AdminUsersTable extends Component {
   @tracked
   users = [];
 
-  sortedBy = "";
+  sortedBy = "username";
 
   constructor() {
     super(...arguments);
@@ -80,8 +80,8 @@ export default class AdminUsersTable extends Component {
             } else if (userB.lastLoginAt == null) {
               return -1;
             } else {
-              if (userA.lastLoginAt < userB.lastLoginAt) return -1 * asc;
-              if (userA.lastLoginAt > userB.lastLoginAt) return 1 * asc;
+              if (userA.lastLoginAt > userB.lastLoginAt) return -1 * asc;
+              if (userA.lastLoginAt < userB.lastLoginAt) return 1 * asc;
               return 0;
             }
           case "last_login_from":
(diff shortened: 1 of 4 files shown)


frontend/app/components/admin/user/table.js
```diff
@@ -9,26 +9,96 @@ export default class AdminUsersTable extends Component {
...
         return !user.isDeleted;
       })
       .sort((userA, userB) => {
-        if (userA.username < userB.username) return -1;
-        if (userA.username > userB.username) return 1;
-        return 0;
+        switch (attribute.toString()) {
+          case "username":
+            if (userA.username < userB.username) return -1 * asc;
+            if (userA.username > userB.username) return 1 * asc;
```
luca-bruegger, line 55:
```suggestion
            return (userA.username < userB.username ? -1 : 1) * asc;
```

frontend/app/components/admin/user/table.js
```diff
@@ -9,26 +9,96 @@ export default class AdminUsersTable extends Component {
...
+            return 0;
+          case "givenname":
+            if (userA.givenname < userB.givenname) return -1 * asc;
+            if (userA.givenname > userB.givenname) return 1 * asc;
+            return 0;
```
luca-bruegger, line 60: Use ternary if possible

frontend/app/components/admin/user/table.js
```diff
@@ -9,26 +9,96 @@ export default class AdminUsersTable extends Component {
...
+              return 0;
+            } else if (userA.lastLoginAt == null) {
+              return 1;
+            } else if (userB.lastLoginAt == null) {
+              return -1;
+            } else {
+              if (userA.lastLoginAt > userB.lastLoginAt) return -1 * asc;
+              if (userA.lastLoginAt < userB.lastLoginAt) return 1 * asc;
+              return 0;
+            }
```
luca-bruegger, line 72: Create a new method for each case additionally. As it increases readability.

luca-bruegger (changes requested):
Wow, congrats! You've implemented an astonishing algorithm to sort values :rocket: 

I left some notes directly within the code. You're on the right path.

Finally, let @Robin481 take a last look, to make sure everything is ok.

RamonaChristen:
Thank you for your feedback! We tried to implement all your suggested changes and everything seems to be working out well. 
@Robin481 do you have anything to add to this?

Robin481:
Hi @RamonaChristen 
This looks really good to me.

2 small notes:
- I don't think the name sorting should consider upper-/lowercase. Examples: Currently the usernames `abc, bcd and cde` will all appear before `Abc, Bcd and Cde`)
This can be solved very easily for example like this:
```          
          case "username":
            return this.sortAttributes(userA.username.toLowerCase(), userB.username.toLowerCase(), asc);
          case "givenname":
            return this.sortAttributes(userA.givenname.toLowerCase(), userB.givenname.toLowerCase(), asc);
```
- Currently the Name sorting (2nd row in the table) doesn't consider a Person's surname so for example several people with the same givenname won't necessarily be sorted correctly when looking at their surname. This should also be easily solved by just giving the sorting "algorithm" the person's fullname instead of only the givenname.

Besides that I really like what you did here, good job!
Implement the little changes from above, make sure the branch is rebased to the current state of the main branch and then there is nothing in the way of a fun little merge :partying_face: :balloon:

Robin481 (approved):
Nice! :tada:

PR #576, opened 2022-03-18
Feature/157 change dialog if pw is empty

MuStAgEDelta:
I add in the translation(de_be,de,en,fr) a phrase for non password. Add test case for non Password and Added tests for a second language(fr) and I changed the Source for the Image in the ReadmeFile, because the old pfad does not exist anymore.
I used the word "mot de passe" as password. I asked a native speaker and he recommended "mot de passe" and not "mot d'accès".


luca-bruegger (changes requested):
So far so good! Two small things to investigate:
- Isn't it better to name the key `none` instead of `non`?
- Move the testcases out from the newly created frontend test file, into the already existing one. I don't think that it is best practice to add a `_fr` suffix into the file name.

After that, let @Robin481 take a closer look at it. Just to be sure that everything is ok.

README.md
```diff
@@ -1,4 +1,4 @@
-![logo](https://github.com/puzzle/cryptopus/blob/master/app/webpacker/images/cryptopussy.svg)
+![logo](https://raw.githubusercontent.com/puzzle/cryptopus/57f8ad8de410e4a0ba16227620727787f22c7d1c/frontend/public/assets/images/cryptopussy.svg)
```
Robin481, line 1: While not in scope of this pull request, still highly appreciated as it annoyed me for the last few weeks :grin:

frontend/tests/integration/components/password-strength-meter-test.js
```diff
@@ -3,7 +3,6 @@ import { setupRenderingTest } from "ember-qunit";
 import { render } from "@ember/test-helpers";
 import { hbs } from "ember-cli-htmlbars";
 import { setLocale } from "ember-intl/test-support";
-
```
Robin481, line 6: Make sure to check your pull requests for random unintended leftovers, I don't think deleting that line here makes sense.

Robin481 (changes requested):
To add on to the comment of Luca:

While I appreciate your effort, there's no point in writing a completely new test for the french translation.
Think about it this way: The only difference between the standard test that uses English translations and your new test using French translations test is the different translations being applied.
Henceforth, the only value you are now adding is that we are testing whether EmberJS correctly applies the translations. As we are developing Cryptopus and not EmberJS it isn't really our task to write tests that cover EmberJS features. 
I would suggest you just delete your newly written test again as it adds unnecessary development overhead for no real advantage on our side.

Another thing on a visual note:
As you can see the now longer sentence that is added causes a line break and kind of ruins the formatting. To that the following suggestion: 
Instead of displaying `Passwort Stärke: Es ist kein Passwort gesetzt`
Display: `Es ist kein Passwort gesetzt` as long as there is no password set. (so basically leave out the `password_strength` part of the string as long as there is no password)
And then just keep it at `Passwort Stärke: Schwach` etc. when there is a password.
This seems like a simple solution that doesn't take any function away, however I am also open to even more creative solutions if you have one :grin: 
This could be done rather easily by adding an if statement to the password_strength translation here: https://github.com/puzzle/cryptopus/blob/9ce26ade9ca26c29e73173caff623ab1af8b6e3a/frontend/app/templates/components/password-strength-meter.hbs#L1

Besides that everything is looking good.
Implement the above changes and rebase to the main branch and we can then merge :partying_face: 

P.S: Make sure to always assign someone so you/me/Luca/anyone else knows whether they need to do something here.

MuStAgEDelta:
@luca-bruegger Thanks for the feedback, I made the changes and we wrote in the other Chat 
@Robin481  thanks for the feedback, I made the changes. To the tests: I understand the feedback. My only thought was, if someone changes something in the translation, it will not be checked. But I understand that testing this for multiple languages would be tedious. 

PS. Yes, I know the readMe File wasn't in my Issue. I think

Robin481 (approved):
Perfect :partying_face: :1st_place_medal:

PR #577, opened 2022-03-18
Add redirect from accounts to encryptables and add a spec

lkleisa:


spec/unit/url_handler_spec.rb
```diff
@@ -23,8 +23,8 @@
   end
 
   it 'return redirect url without accounts when accounts in url' do
-    url = '/de/teams/1/groups/1/accounts/1'
-    url_without_accounts = '/accounts/1'
+    url = 'teams/1/groups/1/accounts/1'
```
luca-bruegger, line 26: `de` prefix still needs to be present

luca-bruegger (changes requested):

Robin481 (approved):
Nice :1st_place_medal:

Issue #370, opened 2021-04-22
Scroll to folder via navigation


TheWalkingLeek:
When activating a folder via navigation, it would be useful to have the folder getting scrolled into the view

mtnstar:
@Robin481 this did not work in my browser somehow.
could u test it too and give feedback to @luca-bruegger ?

PR #581, opened 2022-03-24
Scroll to folder, refs: #370

luca-bruegger:


Robin481 (approved):
Works fine for me.
@mtnstar which Browser were you on?

What I am thinking now however is if we should sticky the left-side navigation bar as it moves out of frame quite quickly... maybe something for the next sprint.

Robin481:
As discussed please check this functionality on Firefox

PR #587, opened 2022-04-01
Add papertrail gem

Drinion:


.rubocop.yml
```diff
@@ -83,6 +83,9 @@ Naming/FileName:
 Rails/EnumHash:
   Enabled: false
 
+Rails/SkipsModelValidations:
```
Drinion, line 86: @RamonaChristen This is only temporary because I think we'll have to find a way to comply with this rule. I invoke `touch` on the encryptable object in the `encryptables_spec` file. Rubocop doesn't like that at all, because invoking touch seems to skip the `encryptable` models validations.

Robin481: Just a little idea: You can also just skip the Rubocop validation for the single spec file if this will be the only place in which this will be used.

Drinion:
TODO: We have implemented the basic functionality of displaying the `Papertrail-Versions`. There are some bugs which we need to get rid off and it may take us about 6 hours to fix them. We also need to implement the functionality to log if someone has `view/copied` a credentials/password.

RamonaChristen:
@Robin481 We are done with the implementation of the read and update log. The failing system specs also fail on the main branch and are not caused by our implementation. Please take a look and let us know if you have any feedback. Thank you!

app/controllers/api/encryptables_controller.rb
```diff
@@ -17,6 +19,8 @@ def index(options = {})
   # GET /api/encryptables/:id
   def show
     authorize encryptable
+    encryptable.paper_trail_event = 'viewed'
+    encryptable.touch
```
Robin481, line 23: Use `encryptable.paper_trail.save_with_version` here.
This will save a version without updating the encryptable's `updated_at`.
Updating the model's `updated_at` every time might have undesirable side effect such as models appearing higher in lists that sort by `updated_at` while not actually having been updated.

app/controllers/api/logs_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  def index(options = {})
+    authorize(team, :index?, policy_class: LogPolicy)
+    render({ json: fetch_entries,
+             each_serializer: list_serializer,
+             root: 'Logs_'.pluralize }
+             .merge(render_options)
+             .merge(options.fetch(:render_options, {})))
+  end
+
+  def fetch_entries
```
Robin481, line 15: I am putting this comment here as this seems like the place where you would solve this problem.
So as we create an Event every time someone opens a encryptable we will (obviously) create **a lot** of events (to test this i created a few thousand events by hand).
It however looks like you are not paginating the results, in turn, loading these thousands events every time someone opens an encryptable.
This will obviously hinder performance and is therefore not advisable. 
The issue states: 
``` 
Find a solution for only showing some of the present log entries if there's a lot
- pagination
- or infinite scroll
- or only display last 20 entries (sql limit 20)
```
Please implement that as well. (a simple solution as limiting the query here to the `.last(20)` or similar should do)

db/migrate/20220401130939_create_versions.rb
```diff
@@ -0,0 +1,38 @@
...
+      # add fractional seconds precision to this migration, specifically, to
+      # the `created_at` column.
+      # (https://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html)
+      #
+      # MySQL users should also upgrade to at least rails 4.2, which is the first
+      # version of ActiveRecord with support for fractional seconds in MySQL.
+      # (https://github.com/rails/rails/pull/14359)
+      # 
+      # MySQL users should use the following line for `created_at`
+      # t.datetime :created_at, limit: 6
```
Robin481, line 33: We (for now) use MySQL in Production so this comment should probably be taken seriously :)

Robin481 (changes requested):
Heya

First of all: Good job and perseverance! :+1: 
Some nice stuff here, the implementation seems to work quite well however it still needs some polishing.
I have commented 2 code specific things already.

Additonally:
You haven't implemented the deletion of Events when a Encryptable is deleted.
You can test this by adding a few versions to an Encryptable by hand
Something like:
```
1000.times do 
  encryptable.touch
end
```
Then deleting the encryptable `encrypable.destroy`
And then calling `PaperTrail::Version.all` to check whether the 1000 versions you just created are still here :grin: 
(hint: `dependent_destroy` is probably your friend here)

Furthermore, this from the issue `Refresh logs whenever an update on an encryptable occurs.` also doesn't seem implemented yet. However, I also think this has the lowest priority besides the other things I have commented.

This is just a quick review from my side with the idea being that you can put some time in on working on the feedback this afternoon (I think you mentioned that you will be working on it this afternoon). There will surely be more code-based feedback but for now this feedback should allow you to bring the feature more towards what it should be.

Cheers and thanks already for the cool work! :partying_face:

app/controllers/api/encryptables_controller.rb
RamonaChristen: Unfortunately, save_with_version does not quite work in the same way as .touch so we have to manually change the event and timestamp of the version like this
```
v = encryptable.versions.last 
v.event = 'viewed'
v.created_at = DateTime.now
v.save
```

which does not look great but it works

app/controllers/api/logs_controller.rb
RamonaChristen: I created a simplified "load more" button which loads 10 logs at a time. To avoid having to get the total amount of logs from the backend most of the logic is implemented in the frontend. So, the button is shown whenever the amount of requested logs from the backend is smaller than or equal to the number of logs displayed. This means that when the number of logs is divisible by 10 the button can be pressed one more time (without anything happening) before it disappears. If that solution doesn't work for you we can also just use the previous commit where simply 20 logs are loaded, without an option to see more.

app/controllers/api/encryptables_controller.rb
Robin481: hmm yes that's a bit clunky.
However a few simple things to make it more okay-ish. Maybe we'll get to a better solution later but for now: 
```
encryptable.paper_trail_event = 'viewed' # we can set that beforehand
v = encryptable.paper_trail.save_with_version # as the method returns the version object
v.created_at = DateTime.now
v.save
```
That still isn't really pretty but it atleast doesn't run the risk of getting the wrong event.

app/models/encryptable.rb
```diff
@@ -17,7 +17,9 @@
 
 class Encryptable < ApplicationRecord
 
-  has_paper_trail on: [:touch, :update], ignore: [:tag, :type, :encrypted_data]
+  has_paper_trail on: [:touch, :update], ignore: [:tag, :type, :encrypted_data], dependent: :destroy
+
+  before_destroy :destroy_versions
```
Robin481, line 22: Why is that necessary? (not necessarily criticizing just trying to understand)

app/serializers/logs_serializer.rb
```diff
@@ -10,7 +10,7 @@ def user_id
   end
 
   def username
-    User.find_by(id: object.whodunnit).username
+    User.find_by(id: object.whodunnit.to_i).username
```
Robin481, line 13: `User.find(object.whodunnit).username`
does the exact same thing :) 
`.find()` is the default by id search.

app/controllers/api/logs_controller.rb
Robin481: That seems like a really nice solution for now, thanks! :car: :dash:

app/models/encryptable.rb
Drinion: Hi @Robin481 The papertrail versions are not deleted from the db by default, if the encryptable they correspond to is deleted. We discussed if it would make sense to keep the old versions because the api user may still want to know what happened to an encryptable that doesn't exist anymore. However, keeping the old encryptables may pollute the db quite heavily. What do you think?

Robin481: Yes deleting them makes sense and was also discussed in #587 .
However, `dependent: :destroy` should already do that, that is why I am asking. :grin: 
They probably implemented that already in #587 so that is why you have the `dependent: :destroy` line here already which should already cover deleting the old versions and therefore making your changes redundant.

Drinion: @Robin481 Yes, `dependent: :destroy` should do that but it doesn't. There might be something wrong with the way used active record with regards to `paper_trail`. I look into it.

app/controllers/api/logs_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+             root: 'Logs_'.pluralize }
+      .merge(render_options)
+      .merge(options.fetch(:render_options, {})))
+  end
+
+  def fetch_entries
+    PaperTrail.serializer = JSON
+    logs = current_user.encryptables.find_by!(id: params[:encryptable_id]).versions
+    if params[:load]
+      logs.last(params[:load]).sort { |a, b| b.created_at <=> a.created_at }
```
mtnstar, line 19: do this in SQL if possible ... 
something like `order(created_at: :desc)`

app/controllers/api/logs_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    else
+      logs.last(10).sort { |a, b| b.created_at <=> a.created_at }
+    end
+  end
+
+  def list_serializer
```
mtnstar, line 25: do not override this method here.
just define used serializer directly on line 8
do not use constantize ... just declare class method directly

app/policies/log_policy.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class LogPolicy < TeamDependantPolicy
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def index?
+    @user.is_a?(User::Human) && team.teammember?(@user.id)
```
mtnstar, line 10: the method 'team_member?' provided by TeamDepdendantPolicy does exactly this.
so just use it :wink:

app/policies/log_policy.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def show?
+    team.teammember?(@user.id)
```
mtnstar, line 14: do we have to override default behavior from TeamDependantPolicy here?

app/policies/log_policy.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  protected
+
+  def team
+    @record
```
mtnstar, line 20: what represents record here?
shouldn't this somehow point from the log entry to the team through associated encryptable ?

app/serializers/logs_serializer.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class LogsSerializer < ApplicationSerializer
+  attributes :id, :item_type, :event, :user_id, :username, :object, :created_at
```
mtnstar, line 4:
```suggestion
  attributes :id, :item_type, :event, :user_id, :username, :created_at
```
do not serialize the whole object here. specify what attributes you require.

app/serializers/logs_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  belongs_to :encryptable
```
mtnstar, line 6: make sure to use `include` where you are fetching the version entries to avoid n+1 queries.

config/routes/api.rb
```diff
@@ -13,6 +13,7 @@
 
     resources :encryptables, except: [:new, :edit] do
       resources :file_entries, only: [:create, :index, :destroy, :show]
+      resources :logs
```
mtnstar, line 16:
```suggestion
      resources :logs, only: [:index]
```
do we need :show too?

spec/models/encryptable_spec.rb
```diff
@@ -84,6 +82,36 @@
...
+      it 'creates a log entry' do
+        encryptable.touch
+        expect(encryptable.versions.count).to eq(1)
+      end
+
+      it 'contains user in whodunnit' do
+        PaperTrail.request.whodunnit = 'Bob'
+
+        encryptable.touch
+        expect(encryptable.versions.last.whodunnit).to eq('Bob')
```
mtnstar, line 110: hmm ... further up `whodunit` included a foreign key (User)
now it includes the username as string.
shouldn't this be always a foreign_key?
please check how this should be done correctly

mtnstar (changes requested):
thank you for the awesome work! looks like you invested quite  a lot of time to create this feature.

some inputs in this review:

- check policy and also create a policy spec
- make sure whodunit is used correctly (user_id)
- make sure there's no n+1 queries

app/policies/log_policy.rb
Drinion: I think we have overridden this method because we also want to allow `Api users` to access the logs. Should we change that?

app/policies/log_policy.rb
Drinion: `@record` is a `Team` object. The way I understand it is that `Team` already represents the `Team` where the `encryptable` is in.

spec/models/encryptable_spec.rb
Drinion: For testing purposes, `whodunnit` is set to a `string` in the `PaperTrail` documentation.

Drinion:
@mtnstar The PR is ready for a Re-review.

mtnstar:
> @mtnstar The PR is ready for a Re-review.

having a look now :wink:

app/controllers/api/encryptables_controller.rb
```diff
@@ -121,4 +124,12 @@ def permitted_attrs
       permitted_attrs
     end
   end
+
+  def log_read_access
+    encryptable.paper_trail.save_with_version
+    v = encryptable.versions.first
+    v.event = 'viewed'
```
mtnstar, line 131:
```suggestion
    v.event = :viewed
```

possible to use symbol here?

app/controllers/api/encryptables_controller.rb
```diff
@@ -121,4 +124,12 @@ def permitted_attrs
       permitted_attrs
     end
   end
+
+  def log_read_access
+    encryptable.paper_trail.save_with_version
+    v = encryptable.versions.first
```
mtnstar, line 130: relying on first here does not sound very safe, does it?

this also creates a complete Version entry with the whole object every time an encryptable is being red.

can't we create a manual entry here for read access? something like:
```
PaperTrail::Version.create!(item_id: encryptable_id, item_type. .... event: :viewed, whodunnit ....)
```

app/controllers/api/logs_controller.rb
mtnstar: load those directly from PaperTrail::Version model
```
PaperTrail::Version.where(item_type: Encrytable.sti_name, item_id: .... ) order bla bli blu
```

app/policies/log_policy.rb
mtnstar: for now, Api Users should not have access to logs.
their read access should be logged though

app/policies/log_policy.rb
mtnstar: @record should be either a PaperTrail::Version entry or a list of those.
so tell the policy how to reach team here. usally the team associated to the encryptable

app/serializers/logs_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+  attributes :id, :item_type, :event, :user_id, :username, :created_at
+
+  belongs_to :encryptable
+
+  def user_id
+    object.whodunnit
+  end
+
+  def username
+    User.find(object.whodunnit).username
```
mtnstar, line 13: this creates also n+1 queries if user is not included in list query

spec/models/encryptable_spec.rb
mtnstar: what are we testing here?
so far, read access seems to be logged inside the encryptables controller.
I think we can trust paper trail to set whodunnit correctly when request.whodunnit is set.
maybe this should be rather be tested in encryptables controller spec #show action

mtnstar (changes requested):
looking better and better ... almost there I think! :rocket: :partying_face: 

* create a test case in encryptables controller spec to make sure read access is being logged
* optimize queries by using arel/sql instead of ruby land
* Rename english versions of table headers: Action, User, Date/Time

app/controllers/api/encryptables_controller.rb
Robin481: I've touched upon this above already (https://github.com/puzzle/cryptopus/pull/587#discussion_r862608971), it just hasn't been implemented yet.
`encryptable.paper_trail.save_with_version` already returns the PaperTrail::Version thingy so we can just do this:
`v = encryptable.paper_trail.save_with_version`

app/serializers/logs_serializer.rb
RamonaChristen: There is no association between Users and Versions therefore I don't think we can use include the username with the Versions query. Is there another way to avoid this n+1 query?

RamonaChristen: I found a way to do it by using join on the user table in the log query :raised_hands:

mtnstar: you should probably use *includes* over joins in that case.
see https://medium.com/@swapnilggourshete/rails-includes-vs-joins-9bf3a8ada00

app/models/encryptable.rb
Robin481: Update on this front?

Robin481 (changes requested):
Eyy, sorry a bit late on that one :clock1: 
I think this should be good now, only the mentioned thread is not resolved yet, any updates there?

RamonaChristen:
done :tada:

PR #588, opened 2022-04-01
533 tech systemspec for folder based actions

renato-oester:


Robin481:
Please fix the Rubocop failures before I have a look at this :giraffe:

renato-oester:
rubocop failures fixed

spec/system/folder_action_system_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+require 'spec_helper'
+
+describe 'FolderAction', type: :system, js: true do
+  include SystemHelpers
+  # can select folder and team from side-navbar
+  it 'opens folder and team in side-navbar' do
+    login_as_user(:admin)
+    visit('/')
+
+    within(page.find('#side-bar-team-235930340')) do
```
Robin481, line 12: So, [Magic Numbers](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad) are never really a good thing. 
In this case I read through the test and do not really have a clue what is going on with this crazy number.
Now, granted there are other examples of this in our test suite so you shouldn't feel too bad for doing it and it's certainly not the end of the world. 
Another thought I had was that the fixture might change that very specific ID at some point in the future but I found out that ActiveRecord Fixtures IDs are apparently [static](https://api.rubyonrails.org/v3.1/classes/ActiveRecord/Fixtures.html#:~:text=The%20generated%20ID%20for%20a,as%20we%20know%20the%20label.) so by this regard this is probably fine. 
Nevertheless, this can easily be evaded by some string interpolation making the test more readable as well in the process -> ex.: `within(page.find("#side-bar-team-#{Team.first.id}))` this works but could be made even more readable by getting the team straight from the logged-in user. Since, from what I can gather the user only has one team so just calling `Team.first` works in this case but also isn't really specific and what if the first team changes in the future? (this is left as an experiment to the reader :stuck_out_tongue_winking_eye:  ).

spec/system/folder_action_system_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      expect(page).to have_text('team1')
+    end
+
+    find('#side-bar-team-235930340').click
+
+    within(page.first('.list-group-item.list-folder-item.bg-blue-one')) do
```
Robin481, line 18: While this works I don't think being that specific here is necessary.
```
within(page.first('.team-card-header')) do
    expect(page).to have_text('folder1')  
end
```
This works just as well and is also more readable. :+1: 

I would suggest you go over all these selection-queries and make them 'less complicated' wherever possible. (Just make sure that we don't break functionality by trying to make things easier. Functionality > Readability)

I tested these statements on my machine by setting a `binding.pry` breakpoint (you might need to add the `pry` gem to your `:test` group in your Gemfile) before said statement and running the test with `HEADLESS=false` and then just inspecting the css classes etc. in the Browser Inspector and then testing my new selection query in the console. :video_game:

spec/system/folder_action_system_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    logout
+
+  end
+
+  # can select folder from team card
+  it 'opens folder in team card' do
+    login_as_user(:admin)
+    visit('/teams/235930340')
```
Robin481, line 35: Again, same thing as above, call the teams ID directly and use String interpolation to make it more readable.

Robin481 (changes requested):
Nice tests :ballot_box_with_check: 
Some small tweaks in the review :package:

spec/system/folder_action_system_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  # Run before each testcase
+  before(:each) do
+    login_as_user(:admin)
+    @encryptable = Encryptable.first
+    @folder = Folder.find(@encryptable.folder_id)
+    @team = Team.find(@folder.team_id)
+  end
+
+  # can select folder and team from side-navbar
+  it 'opens folder and team in side-navbar' do
```
Robin481, line 17: Now about our discussion from Tuesday about necessary or non-necessary comments :joy: 
You could just name the test case as to make the comment unnecessary e.g. `it "can select folders and teams from sidebar"`
However, I digress and there is no need to spend more time on such small things for a test case :grin: 
Everything else looks good to me, good job!

Robin481 (approved):

PR #613, opened 2022-05-12
598 tech add support for postgresql and pgsearch

Drinion:


Drinion:
Hello @Robin481 We are currently a little stuck with the tech upgrade. Could you maybe have a look at the code to see, if it's going into the right direction? Thanks a lot!

config/database.yml
```diff
@@ -6,16 +6,14 @@
 # PostgreSQL. Versions 8.2 and up are supported.
 #
 base: &generic
-  adapter: <%= ENV['RAILS_DB_ADAPTER'] || 'sqlite3' %>
+  adapter: sqlite3
```
Robin481, line 9: Why make `sqlite3` the default here.
You want the opposite.
Either make PostgreSQL the default or leave it as it was and set the environment variable `RAILS_DB_ADAPTER`.

config/docker/mysql/docker-compose.yml
```diff
@@ -1,28 +1,11 @@
-version: '2'
-
-services:
```
Robin481, line 3: You can probably just delete the whole mysql service, or just leave it in but don't use it.

config/docker/postgresql-development/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '3'
+
+services:
+  postgresql:
+    image: postgres
```
Robin481, line 5: Looks good, (I haven't checked how much works and how much doesn't).
Is this a proper image reference or do you have to be more specific?
You then have to set the proper ENV variables so that Rails knows where to find the PostgreSQL container  in the `database.yml` file. (you set the port number here so for example so you can just give rails that port number etc.)

Robin481: Also I would start by just making sure that the PostgreSQL container works.
- Start the postgresql container with docker and see if you have any errors starting.
- If it starts connect via `docker exec -it NAME_OF_CONTAINER bash` and try to connect to the postgresql instance from there.
- If you know that all works you can connect rails to the postgresql container and rails should then be able to create the database, migrate and see it (famous last words)
- To be even more thorough you can then connect to the postgresql container again and connect to the newly created database and see if the tables that rails just created are there and are functioning properly (they really should be there but I think this helps understanding :grin: )

db/schema.rb
```diff
@@ -98,4 +98,14 @@
     t.index ["default_ccli_user_id"], name: "index_users_on_default_ccli_user_id"
   end
 
+  create_table "versions", force: :cascade do |t|
```
Robin481, line 101: I am not entirely sure why you have the versions table here.
Probably a left over from working on the other branches?
As you are basing this branch on master which currently doesn't have any versions this would probably break some stuff :joy: unless I am misunderstanding something :angel:

Robin481 (commented):
Heya looks good so far. 
I left a comment that should help guide you towards a working PostgreSQL container.
If you need more specific help let me know what you are struggling with.

db/schema.rb
Drinion: Whoops! yes this is a mistake 😆

PR #614, opened 2022-05-13
Improvment/translation encryptable credentials

lkleisa:
Before review rebase needed @mtnstar


app/models/encryptable/file.rb
```diff
@@ -12,6 +12,10 @@ class Encryptable::File < Encryptable
 
   validate :file_size, on: [:create, :update]
 
+  validate :file_size, on: [:create, :update]
+
+  validate :file_size, on: [:create, :update]
```
mtnstar, line 17: duplicate statement ?

frontend/app/templates/components/nav-bar.hbs
```diff
@@ -38,7 +38,7 @@
       <dd.menu @align="right" as |ddm|>
         <ddm.item>
           <a role="button" class="dropdown-item" {{on "click" this.toggleEncryptableCreating}}>
-            {{t "encryptables.new.title"}}
+            {{t "encryptable/credential.new.title"}}
```
mtnstar, line 41:
```suggestion
            {{t "encryptable/credentials.new.title"}}
```
it's a bit confusing, I know ... but we use the term encryptable/credentials (plural)
since an encryptable/credentials can include multiple credentials (username, password, ...)
so you also say ... please enter your credentials to login in ...

mtnstar (changes requested):
looks pretty goooood ... make sure to do manual testing in UI ...

please use encryptable/credentials, not encryptable/credential for Encryptable::Credentials

thx

frontend/app/templates/components/nav-bar.hbs
lkleisa: Yes, this makes sense! I changed it in the latest commit

app/models/encryptable/file.rb
lkleisa: Removed with rebasing

PR #620, opened 2022-05-17
383 ccli login command button

renato-oester:


frontend/app/components/nav-bar.js
```diff
@@ -66,4 +70,22 @@ export default class NavBarComponent extends Component {
...
+
+  @action
+  copyCcliLogin() {
+    this.fetchService
+      .send(`/api/api_users/ccli_token`, { method: "GET" })
+      .then((response) => {
+        response.json().then((json) => {
+          this.clipboardService.copy(
+            `cry login ${btoa(`${json.username}:${json.token}`)}@${
+              window.location.origin
```
Robin481, line 82: I am wondering whether it would make more sense to create the whole string in the backend and use backend ENV variables to do this :thinking: 
`ENV['RAILS_HOST_NAME'] ` would probably fit best. (I checked our Openshift environment and it is correctly set to the relevant url.)
I'll let @mtnstar chime in here :grin:

spec/system/nav_bar_system_spec.rb
```diff
@@ -45,4 +45,22 @@
...
+      dropdown_menus.second.click
+      expect(find('pzsh-menu-dropdown-item', text: 'Copy CCLI Login')).to be_present
+    end
+
+    it 'copies ccli login command to clipboard' do
+      login_as_user(:admin)
+      dropdown_menus = all('pzsh-menu-dropdown', text: users(:admin).givenname).to_a
+      dropdown_menus.second.click
+      find('pzsh-menu-dropdown-item', text: 'Copy CCLI Login').click
+      expect(find('span.message', text: 'CCLI Login command was copied!')).to be_present
```
Robin481, line 62: Have you checked if there is a way to see what was copied to the clipboard?
This way we only check if we have a success message however we don't check the clipboard functionality.
Would be nice to have :smile_cat:

Robin481 (changes requested):
This seems good to me, one little question I have that I wan't @mtnstar 's input for and a little test note but besides that I don't see any problems. 
Good job! :1st_place_medal:

spec/system/nav_bar_system_spec.rb
renato-oester: Unfortunately i can't figure out a way to read the clipboard content.
I have looked at the clipboard API, but i can't get that to work.
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

Robin481: Alright that's fine, was worth a shot :grin:

frontend/app/components/nav-bar.js
mtnstar: I would leave it as it is for the moment. 
not so sure if it's a good idea to rely on RAILS_HOST_NAME env

Robin481 (approved):

Robin481:
Good Job :partying_face:

PR #624, opened 2022-05-23
Bug/618/favorite team duplication

njaeggi:


app/models/team/personal.rb
```diff
@@ -21,6 +21,8 @@ def create(owner)
       team = super(name: owner.username, personal_owner: owner, private: true)
       return team unless team.valid?
 
+      UserFavouriteTeam.create(team_id: team.id, user_id: owner.id)
```
mtnstar, line 24: do not create this entry. just assume that personal team is always favourised

app/serializers/team_serializer.rb
```diff
@@ -21,7 +21,7 @@ class TeamSerializer < ApplicationSerializer
   has_many :folders, serializer: FolderMinimalSerializer
 
   def favourised
-    users_favorised_current_team.include?(user.id)
+    user_favourite_team_ids.include?(object.id)
```
mtnstar, line 24:
```suggestion
    object.personal_team? || user_favourite_team_ids.include?(object.id)
```

app/serializers/team_serializer.rb
```diff
@@ -38,7 +38,7 @@ def user
     current_user
   end
 
-  def users_favorised_current_team
-    @users_favorised_current_team ||= object.user_favourite_teams.pluck(:team_id)
+  def user_favourite_team_ids
+    @user_favourite_team_ids ||= object.user_favourite_teams.pluck(:team_id)
```
mtnstar, line 42: this does not work ... this return all favourite team entries from all users !
you need to get favourite teams from current user

Issue #560, opened 2022-03-09
ENCRYPABLES SHARING: Credentials share


njaeggi:
Implement the possibility to share credential

### Conception
- https://codimd.puzzle.ch/tozsq7r-STCJ29iqsMbS9w#

The whole implementation of Encryptable Sharing should be done with
- https://github.com/puzzle-bbt/bbt/issues/104
- https://github.com/puzzle/cryptopus/issues/561

The button to share a credential should also be implemented
- https://github.com/puzzle/cryptopus/issues/611

Todo:
- [ ] shares credential when selecting user

PR #636, opened 2022-06-07
Feature file sharing domain class, refs: #560

luca-bruegger:


app/controllers/api/encryptables_controller.rb
```diff
@@ -104,7 +106,7 @@ def tag_param
   end
 
   def encryptable_move_handler
-    EncryptableMoveHandler.new(encryptable, session[:private_key], current_user)
+    EncryptableMoveHandler.new(encryptable, users_private_key, current_user)
```
mtnstar, line 109: :+1:

app/controllers/concerns/encryptables.rb
```diff
@@ -56,6 +56,49 @@ def credential_id
...
+  def transfer_encryptable
+    sender_id = current_user.id
+
+    shared_encryptable = EncryptableTransfer.new.transfer(encryptable, receiver_id, sender_id)
+
+    instance_variable_set(:"@#{ivar_name}", shared_encryptable)
+  end
+
+  def transfered_encryptable?(entry)
+    entry.encrypted_transfer_password.present? && entry.sender_id.present?
```
mtnstar, line 74: maybe add getter to encryptable model ?
#transfered?

app/controllers/concerns/encryptables.rb
```diff
@@ -56,6 +56,49 @@ def credential_id
...
+  end
+
+  def receiver_id
+    params.dig('data', 'attributes', 'receiver_id')
```
mtnstar, line 78: can't we use model_params helper method here?

app/controllers/concerns/encryptables.rb
```diff
@@ -56,6 +56,49 @@ def credential_id
...
+  end
+
+  def decrypt_shared_encryptable(entry, private_key)
+    plaintext_transfer_password = Crypto::Rsa.decrypt(entry.transfer_password, private_key)
+    entry.decrypt(plaintext_transfer_password)
+
+    recrypt_with_personal_team_password(entry)
+    remove_shared_attributes(entry)
```
mtnstar, line 86: remove transfer attributes when doing a recrypt inside encryptable model. 
this makes sure the encryptable entry cannot be in an invalid state (recrypted, but still sender_id and transfer_password set)

app/controllers/concerns/encryptables.rb
```diff
@@ -56,6 +56,49 @@ def credential_id
...
+  def encryptable_transfering?
+    receiver_id.present?
+  end
+
+  def transfer_encryptable
+    sender_id = current_user.id
+
+    shared_encryptable = EncryptableTransfer.new.transfer(encryptable, receiver_id, sender_id)
+
+    instance_variable_set(:"@#{ivar_name}", shared_encryptable)
```
mtnstar, line 70: use static name ... probably @encryptable

app/controllers/concerns/encryptables.rb
```diff
@@ -56,6 +56,49 @@ def credential_id
     params[:credential_id]
   end
 
+  ### Sharing ###
```
mtnstar, line 59: since we're using domain class EncryptableTransfer ... it probably does not require an additional file encryptable concern. 
so I would suggest removing this completely and add some minimal code to EncryptablesContrroller directly

app/policies/encryptable_policy.rb
```diff
@@ -31,4 +35,23 @@ def destroy?
...
+      current_user.present? &&
+      user_human? &&
+      sender_himself?
+  end
+
+  def user_human?
+    User.find(@record.receiver_id).type == User::Human
+  end
+
+  def sender_himself?
```
mtnstar, line 53: do we really have to check this?
is this security related?

frontend/app/components/encryptable/form.js
```diff
@@ -1,5 +1,7 @@
 import { action } from "@ember/object";
-import AccountValidations from "../../validations/encryptable";
+import EncryptableValidations from "../../validations/encryptable";
```
mtnstar, line 2: completely focus on backend for now.
make sure all backend functionality is working ... then implement frontend.
TDD for the win

spec/controllers/api/encryptables_controller_spec.rb
```diff
@@ -700,6 +700,27 @@
     end
   end
 
+  context 'share Encryptable' do
```
mtnstar, line 703:
```suggestion
  context 'encryptable transfer' do
```

spec/controllers/api/encryptables_controller_spec.rb
```diff
@@ -700,6 +700,27 @@
...
+    it 'build_entry returns shared_encryptable if shared_encrytable is true' do
```
mtnstar, line 704:
```suggestion
    it 'sends encryptable to recipient' do
```

mtnstar: try to describe test cases like use cases. so non technical if possible

spec/migrations/move_files_from_credential_to_folder.spec.rb
```diff
@@ -0,0 +1,194 @@
+# frozen_string_literal: true
```
mtnstar, line 1: this file does not belong in this PR ! remove it ...
it was renamed on master !

mtnstar (changes requested):
keep going :rocket: ... some feedback :smile:

config/locales/ch_be.yml
```diff
@@ -96,6 +96,10 @@ ch_be:
           attributes:
             name:
               taken: Datei isch schomau ufägladä wordä
+        encryptable:
+          attributes:
+            name:
+              taken: Du hesch die zuegangsdate scho mit derä person teilt
```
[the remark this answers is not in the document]
mtnstar: we probably need to adjust uniqueness constraint. 
so im sender_id is set, name does not have to be unique inside folder anymore.
or maybe if a encryptable with same filename already exists ... we just add name(1).pdf ...
like a file browser manages this situation.

spec/controllers/api/encryptables_controller_spec.rb
```diff
@@ -700,6 +701,169 @@
     end
   end
 
+  context 'encryptable transfer' do
+    it 'sends encryptable credentials to recipient' do
```
mtnstar, line 705: we really want to focus on sending files for now!

spec/controllers/api/encryptables_controller_spec.rb
```diff
@@ -700,6 +701,169 @@
...
+
+      share_encryptable_params = {
+        data: {
+          attributes: {
+            id: file1.id,
+            receiver_id: 657656756567856
+          }
+        }
+      }
+      expect { post :create, params: share_encryptable_params, xhr: true }.to raise_error(StandardError, "Receiver user not found")
```
mtnstar, line 807: put this with do ... end on multiple lines

mtnstar: this should raise user not found exception by active record.
so no need to implement this yourself. check comments further up

spec/fixtures/files/users/api_tux/public.key
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
```
mtnstar, line 1: what's this for ?

spec/models/user/human_spec.rb
```diff
@@ -52,6 +52,24 @@
         expect(Team.exists?(personal_team_id)).to eq(false)
       end
 
+      it 'deletes personal team when user is deleted' do
```
mtnstar, line 55: does this belong in here?
is it an extension to given personal team feature?

spec/policies/folder_policy_spec.rb
```diff
@@ -32,6 +33,14 @@
     it 'can destroy folder' do
       assert_permit bob, folder2, :destroy?
     end
+
+    it 'cannot destroy folder when in personal-team' do
```
mtnstar, line 37: should be possible to destroy all the folders.
if inbox does not exist, we are creating it

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    let(:encryptable_file) do
+      Encryptable::File.new(cleartext_file: 'fff', name: 'info.txt', content_type: 'text/plain')
+    end
+
+    it 'sends encryptable file to other user' do
+      receiver = bob
+      sender_id = alice.id
+
+      encryptable_share.transfer(encryptable_file, receiver, sender_id)
```
mtnstar, line 27: first use #send method ...

then use #receive method ...

mtnstar (changes requested):
nice nice ! :partying_face: go go :rocket: 

* focus on sending files only for now
* try to move as much transfer logic as possible to transfer domain class
* equip transfer domain class with send and receive methods
* do not raise standard error ... just use default error ... e.g. ActiveRecord user not found error. Write spec and see what error occurs.

PR #641, opened 2022-06-10
Respect line breaks in descriptions

codez:
We often use line breaks in password descriptions. With this change, they are visible in the show view.


MarcEgliP (approved):
Everything works perfectly!

codez:
:tada:

PR #625, opened 2022-05-24
Feature recrypt, refs: #593

luca-bruegger:
So far done. Needs merge (rebase) with master and another review from @mtnstar regarding the frontend changes for the recrypt.

Issue #595, opened 2022-04-05
ENCRYPTABLES: RECRYPT controller


luca-bruegger:
As of today, the recrypt should be triggered on team show. Does one team not have the latest algorithm set and is not being recrypted, then it shall be recrypted. This should be done by wiring-up the new domain class, where the logic is contained.

# ToDo
- [x] Make new Check in `Teams#show` to check if latest algorithm is in use
- [x] If not, recrypt the given team with `Crypto::Recrypt` Class
- [x] If latest algorithm is in use, nothing should be changed

PR #642, opened 2022-06-17
Wire up teams controller to recrypt, refs: #595

luca-bruegger:
Based on #625

### DIFF
--- a/app/controllers/api/teams_controller.rb
+++ b/app/controllers/api/teams_controller.rb
@@ -17,7 +17,7 @@ class Api::TeamsController < ApiController
   # GET /api/teams
   def index
     if params['team_id'].present?
-      authorize fetch_entries.first, :team_member?
+      send_team
     elsif params['only_teammember_user_id'].present?
       authorize ::Team, :only_teammember?
     else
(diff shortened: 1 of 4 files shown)


app/controllers/api/teams_controller.rb
```diff
@@ -28,6 +28,24 @@ def index
 
   private
 
+  def team_response
```
mtnstar, line 31:
```suggestion
  def fetch_team
```

app/controllers/api/teams_controller.rb
```diff
@@ -28,6 +28,24 @@ def index
...
+    team = fetch_entries.first
```
mtnstar, line 32: better @team ?

there's already a getter `def team`, is this still required?
maybe use this getter to cache team ?

db/schema.rb
```diff
@@ -22,7 +22,9 @@
     t.text "encrypted_data", limit: 16777215
     t.integer "credential_id"
     t.text "content_type"
-    t.string "encryption_algorithm"
+    t.binary "transfer_password"
```
mtnstar, line 25: does not belong in here probably ...

db/schema.rb
```diff
@@ -61,7 +63,7 @@
     t.string "type", default: "Team::Shared", null: false
     t.integer "personal_owner_id"
     t.integer "recrypt_state", default: 1, null: false
-    t.string "encryption_algorithm"
+    t.string "encryption_algorithm", default: "AES256", null: false
```
mtnstar, line 66: no default ... make sure you're checking in the right schema.rb
check this file out ... then run db:migrate ... this should create an up to date version

mtnstar (changes requested):
well done ! almost perfect.

Issue #561, opened 2022-03-09
ENCRYPTABLE SHARING: Extend Encryptables Controller for File sharing


njaeggi:
Adjust the post create action to handle an additional receiver id param to share files.

### Conception
- https://codimd.puzzle.ch/tozsq7r-STCJ29iqsMbS9w#

# Todo
- [x] Implement specs for transfer domain class
- [x] Implement review feedback for transfer domain class
  - [x] Create personal Team inbox folder if nonexistent
- [x] Edit EncryptablePolicy create?
  - [x] true when user is logged in and receiver is not API User or the sender himself
- [x] Write new Specs for Sharing Feature in EncryptableControllerSpec
  - [x] Make sure that it is impossible to send to api-users and nonexistent recipients
  - [x] Create two sets of permitted attrs. Use one according situation.
  - [x] Create new flash message if file was sent successfully or unsuccessfully
- [x] Use transfer domain class in Encryptable Controller

PR #646, opened 2022-06-21
Implement file sharing domain class, refs: #561

luca-bruegger:
Domain class for file sharing, as mentioned in #561


app/policies/encryptable_policy.rb
```diff
@@ -15,10 +15,14 @@ def show?
...
   def create?
+    return true if encryptable_transfer
+
     team.teammember?(@user.id)
   end
 
   def update?
+    return true if encryptable_transfer
+
     team.teammember?(@user.id)
```
mtnstar, line 26:
```suggestion
    team.teammember?(@user.id) || transfering_encryptable?
```

app/policies/encryptable_policy.rb
```diff
@@ -31,4 +35,18 @@ def destroy?
   def team
     @record.team
   end
+
+  private
+
+  def encryptable_transfer
+    @record.receiver_id.present? &&
+      user.present? &&
```
mtnstar, line 43: is always present if logged in ... so no need to check

app/policies/encryptable_policy.rb
```diff
@@ -31,4 +35,18 @@ def destroy?
   def team
     @record.team
   end
+
+  private
+
+  def encryptable_transfer
+    @record.receiver_id.present? &&
```
mtnstar, line 42: should we check for team.personal_owner ?

app/policies/encryptable_policy.rb
```diff
@@ -15,10 +15,14 @@ def show?
   end
 
   def create?
+    return true if encryptable_transfer
+
     team.teammember?(@user.id)
   end
 
   def update?
+    return true if encryptable_transfer
```
mtnstar, line 24: why is this also available for update ?

app/utils/encryptable_transfer.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class EncryptableTransfer
+  def transfer(encryptable, receiver, sender_id)
```
mtnstar, line 4:
```suggestion
  def send(encryptable, receiver, sender_id)
```

app/utils/encryptable_transfer.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  private
+
+  def inbox_folder(receiver)
+    personal_team = receiver.personal_team
+    personal_team.folders.find_or_create_by(name: 'inbox')
+  end
+
+  def encrypted_transfer_password(password, receiver)
```
mtnstar, line 37: maybe use the method from encryptable model?

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    it 'does not transfer encryptable file to non existent user' do
+      receiver = nil
+      sender_id = alice.id
+
+      expect do
+        encryptable_transfer.transfer(encryptable_file, receiver,
+                                      sender_id)
+      end.to raise_error(StandardError, 'Receiver user not found')
```
mtnstar, line 49: check for error that happens since it cannot be assigned. 
no need to have a special exception for that case ... probably active record fails to find receiver personal team

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe EncryptableTransfer do
+
+  let(:encryptable) { encryptables(:credentials1) }
```
mtnstar, line 7: still required ?

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  let(:encryptable_file) do
+    Encryptable::File.new(cleartext_file: 'fff', name: 'info.txt', content_type: 'text/plain')
+  end
+
+  context '#transfer' do
+    it 'transfers encryptable file to other user' do
+      receiver = bob
+      sender_id = alice.id
+
+      encryptable_transfer.transfer(encryptable_file, receiver, sender_id)
```
mtnstar, line 25: maybe provide receiver and sender
it would make it more constant.
so not one time only the id and then the whole user.
sender is always current_user ... so we do have the full record available

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  let(:bobs_private_key) { bob.decrypt_private_key('password') }
+  let(:team1) { teams(:team1) }
+
+  let(:encryptable_transfer) { described_class.new }
+
+  let(:encryptable_file) do
+    Encryptable::File.new(cleartext_file: 'fff', name: 'info.txt', content_type: 'text/plain')
+  end
+
+  context '#transfer' do
```
mtnstar, line 20: for me it would be ok just having one context block ... transfer

and then test send and receive in the same block

mtnstar (changes requested):
thanks for the great work! it's getting better and better ...

still some things to check

* decide what parts of this feature's code should go into the domain class and what into the model class
* receiver is assigned through personal team ... so access it that way
* remove duplicated code ... see comments

app/utils/encryptable_transfer.rb
lkleisa: Why do we change now from transfer back to send?

app/controllers/concerns/encryptables.rb
```diff
@@ -51,7 +51,7 @@ def new_file(parent_encryptable, description, name)
   end
```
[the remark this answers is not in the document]
lkleisa: I changed it, but in my view, there's too much code in the encryptables controller

app/models/encryptable.rb
```diff
@@ -35,6 +40,10 @@ def decrypt(_team_password)
     raise 'implement in subclass'
   end
 
+  def decrypt_transfered(_private_key)
+    raise 'implement in subclass'
```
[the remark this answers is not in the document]
lkleisa: In this case it's not possible, because we only want to implement this for Encryptable::File for file transfer

app/models/encryptable.rb
```diff
@@ -27,6 +28,9 @@ class Encryptable < ApplicationRecord
   validates :name, presence: true
   validates :description, length: { maximum: 4000 }
 
+  validate :assert_human_receiver?, if: :transferred?
```
lkleisa, line 31: Currently, there is an error with this validation. It's not called with the test, but it's working in rails console

app/policies/encryptable_policy.rb
```diff
@@ -31,4 +31,22 @@ def destroy?
...
+
+  def transferring_encryptable?
+    @record.receiver_id.present? &&
+      receiver_personal_team_owner? &&
+      user_human? &&
+      @record.is_a?(Encryptable::File)
+  end
+
+  def user_human?
+    User.find(@record.receiver_id).is_a?(User::Human)
```
mtnstar, line 45: combine with `@record.folder.team.owner.id` ?
so we could get rid of attr accessor `receiver_id`

app/utils/encryptable_transfer.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class EncryptableTransfer
+  def transfer(encryptable, receiver, sender)
```
mtnstar, line 4:
```suggestion
  def send(encryptable, receiver, sender)
```

app/utils/encryptable_transfer.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def receive(encryptable, private_key, personal_team_password)
+    encryptable.decrypt_transfered(private_key)
+
+    encryptable.update!(encrypted_transfer_password: nil,
+                        receiver_id: nil,
+                        sender_id: nil)
+
+    encryptable.encrypt(personal_team_password)
+    encryptable.decrypt(personal_team_password)
```
mtnstar, line 25: do we need to encrypt/decrypt here?
we're already doing this at decrypt_transfered?

app/utils/encryptable_transfer.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    encryptable
+  end
+
+  private
+
+  def inbox_folder(receiver)
+    raise StandardError.new "Receiver not set" if receiver.nil?
```
mtnstar, line 33: do we have to check this?
should fail at receiver.personal_team anyways?

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    it 'transfers encryptable file to other user' do
```
mtnstar, line 20: very nice ! :partying_face:

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    it 'does not transfer encryptable file to non existent user' do
+      receiver = nil
+      sender = alice
+
+      expect do
+        encryptable_transfer.transfer(encryptable_file, receiver,
+                                      sender)
+      end.to raise_error(StandardError, 'Receiver not set')
```
mtnstar, line 47: leave this to active record record not found ...

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    it 'does not transfer encryptable file to api user' do
+      receiver = api_tux
+      sender = alice
+
+      expect do
+        encryptable_transfer.transfer(encryptable_file, receiver,
+                                      sender)
+      end.to raise_error(StandardError, 'Cant transfer to API user')
```
mtnstar, line 57: maybe also leave to active record not found ...
not need to have a customized message here

mtnstar (changes requested):
nice! keep going ... almost done!

app/utils/encryptable_transfer.rb
njaeggi: Why do we change now from transfer back to send?

app/utils/encryptable_transfer.rb
```diff
@@ -0,0 +1,48 @@
...
+    )
+    encryptable
+  end
+
+  def receive(encryptable, private_key, personal_team_password)
+    encryptable.decrypt_transfered(private_key)
+
+    encryptable.update!(encrypted_transfer_password: nil,
+                        receiver_id: nil,
+                        sender_id: nil)
```
mtnstar, line 22: we should not reset sender, should we ?

spec/fixtures/users.yml
```diff
@@ -84,3 +84,14 @@ conf_admin:
   public_key: <%= "!!binary \"#{FixturesHelper.read_public_key('conf_admin')}\"" %>
   private_key: <%= "!!binary \"#{FixturesHelper.read_private_key('conf_admin')}\"" %>
 
+api_tux:
```
mtnstar, line 87: why do we need an API fixture user here?
can't we just fabricate one where used in tests?

lkleisa: We noticed, that we don't need it here in domain class, because the tests for api_user and non existent user are made in controller

PR #653, opened 2022-07-05
display flash message when empty value is copied

njaeggi:


njaeggi:
Only works in encryptable/row.js, since we are using different method to copy. The other copy buttons cant copy empty strings.

Issue #657, opened 2022-07-20
ENCRYPTABLE SHARING: Finish full feature 


lkleisa:
We started with the feature for Encryptable::File sharing in these stories: 
Controller: #561 
UI: #611 
This is the branch with the latest version: https://github.com/puzzle/cryptopus/pull/658

### Todo 
- [x] File download from encryptable/row and encryptable/show when clicking the button in row or show view -> File is correctly attached to the encryptable. Only the path for downloading the file is missing. This must be done here: 1. [row.js](https://github.com/puzzle/cryptopus/blob/feature/657/file-sharing/frontend/app/components/encryptable/row.js#L120), 2. [show.js](https://github.com/puzzle/cryptopus/blob/feature/657/file-sharing/frontend/app/components/encryptable/show.js#L42)
- [x] Fix bug when downloading file from encryptable/show and reloading that name is not filled out
- [x] Check for other data types -> The specs and manual tests were made with plain text files (.txt). Image uploads are currently not working, maybe this must be fixed. Check also .pdf files
- [x] Check if the validation for the uploaded file is working (not too big and not empty)
- [x] Fix this spec: [Order of the encryptables](https://github.com/puzzle/cryptopus/blob/feature/657/file-sharing/spec/controllers/api/teams_controller_spec.rb#L182) -> The encryptables in the inbox folder are ordered after created_at. This is working in the browser, but somehow not in this spec (maybe nowhere in the backend...)
- [x] Create component specs for the new UI features
- [x] Update specs in rails for the new features (encryptables_controller)
- [x] Disable rubocop in encryptable.rb:73
- [x] Show Badge on Personal team to show number of unread transfered files "bob (42)"
- [x] Create/check Ticket for credential sharing -> [New Ticket](https://github.com/puzzle/cryptopus/issues/688)
- [x] Fix notification when file transferred successfully
- [x] Fix system specs in pipeline

### Optional
- [ ] Display inbox folder at the top of the users personal team
- [ ] Don't allow user to rename inbox folder (remove is not a problem, because we generate an inbox folder if he isn't existing)

lkleisa:
Badge is working, but only one time, then if you show one transferred file, the request for getting the amount of unread transferred files is not done again because we only fetch one time the teams for the side nav bar.

PR #658, opened 2022-07-20
Feature/657/file sharing

lkleisa:
In progress: #657


app/models/encryptable.rb
```diff
@@ -47,16 +46,37 @@ def team
...
+  end
+
   private
 
   def encrypt_attr(attr, team_password)
     cleartext_value = send(:"cleartext_#{attr}")
 
-    encrypted_value = if cleartext_value.blank?
+    # rubocop:disable all
+    encrypted_value = if cleartext_value.nil? || cleartext_value.empty?
```
lkleisa, line 74: Needs to be done like this because when using .blank? or &.blank? on transfer from images and pdf's, then we get this error:
image

mtnstar:
image

strange shadow when selecting recipient?

mtnstar:
image

badge is good ... but does it have to be red and so dominant?
maybe better vertical centering?

mtnstar:
image

1. should we prevent user from deleting their inbox?
2. should we display also a badge for the unread count on the folder inbox? maybe even if count is 0?
3. Is it clear for user that the :person_curly_hair: means sender? Sent by? better to have text? or at least a tooltip?

frontend/app/templates/components/file-transfer.hbs
```diff
@@ -0,0 +1,69 @@
...
+              <ValidationErrorsList @errors={{this.changeset.error.description.validation}}/>
+            </div>
+          </div>
+        </div>
+      </BsForm>
+    </div>
+  </Modal.body>
+
+  <Modal.footer>
+    <BsButton @onClick={{action this.submit this.changeset}} alt="save" @type="primary">{{t "encryptable/transferring/file.submit"}}</BsButton>
```
mtnstar, line 67: maybe disable button unless all required fields/data is present?

spec/fixtures/encryptables.yml
```diff
@@ -41,6 +40,14 @@ file1:
   content_type: 'text/plain'
   encrypted_data: '{"file":{"iv":null,"data":"FvJS9jooGEX1aXqB0iP7wB6h2YwO479OM+RpNmBlbORivbVPky0rR4u3lNLN+DGIL93gQAlVHDw1CZe9zDoTgSyxsQFflQwGk3DMDS9xhoQSJzTkBPBIb33j9H7WG37CQwdNNFnn/NExiBZb+9dbmHGqw8KWvRd3Xd/oSlTr6w/c0gz3UEYfhC5l6P3xnDc2"}}'
 
+transferredFile1:
```
mtnstar, line 43: ruby world does not use camelCase ... use snake_case please :wink:

spec/unit/utils/encryptable_move_handler_spec.rb
```diff
@@ -32,6 +32,15 @@
 
     expect(folders(:folder2).id).to eq(credential.folder_id)
 
+    new_encryptable_file = Encryptable::File.new(name: 'testFile',
+                                                 cleartext_file: file_fixture('test_file.txt').read,
+                                                 credential_id: credential.id,
+                                                 content_type: 'text/plain')
+
+    team2_password = Base64.strict_decode64('Xyj5d0yF9D/XOCIi9Iz5bsgNs9KvvcKkJAtCsoENNj4=')
```
mtnstar, line 40: isn't there a global helper available for this?

mtnstar (changes requested):
wow, it's an impressive and huge feature !!! well done so far :rocket: 

some things to check and improve, see comments

app/controllers/api/encryptables_controller.rb
```diff
@@ -18,34 +17,37 @@ def index
   # GET /api/encryptables/:id
   def show
     authorize entry
-    entry.decrypt(decrypted_team_password(team))
+
+    if entry.transferred?
+      personal_team = current_user.personal_team
+      personal_team_password = decrypted_team_password(personal_team)
+      EncryptableTransfer.new.receive(entry, session[:private_key], personal_team_password)
```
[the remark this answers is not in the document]
lkleisa: In our request to show a transferred encryptable, we don't send `Authorization-Password` in the headers. The method users_private_key uses these headers, and this is here nil. Because we use `session[:private_key]` multiple places, it should be okey like that, right?

db/migrate/20220719120958_add_personal_inbox_to_folders.rb
```diff
@@ -0,0 +1,14 @@
...
+    Team::Personal.find_each do |t|
+      folder = t.folders.find_by(name: 'inbox')
+      folder&.update!(personal_inbox: true)
+    end
+  end
+
+  def down
+    remove_column :folders, :personal_inbox, :boolean
+  end
+end
```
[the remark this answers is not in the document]
lkleisa: Model validation in folder.rb:23 should do this, doesn't it?
` validates :name, uniqueness: { scope: :team }`

lkleisa:
Feedback from PR is implemented, App is working as it should. System specs are locally running, but not in pipepline

kcinay055679:
grafik
Shareicon is not displayed

kcinay055679:
grafik
Cryptopus thows error after uploading a zip file

app/policies/encryptable_policy.rb
```diff
@@ -31,4 +33,11 @@ def destroy?
   def team
     @record.team
   end
+
+  private
+
+  def transferring_encryptable?
```
njaeggi, line 39: maybe rename this method to receiving_encryptable, as I understand this method allows encryptables to be created when receiving one, and not for transfering (sending), but I might just misunderstand the context

config/locales/ch_be.yml
```diff
@@ -81,6 +81,8 @@ ch_be:
       weak_password: 'Um d Sicherheit z verbessere, söttisch es komplizierteres Passwort wähle'
     teams:
       not_existing: 'Es existiert kes Team mit dr Id %{id}'
+    folders:
+      duplicated_inbox: 'Es cha nume ei inbox ordner im personal team gä'
```
njaeggi, line 85: Maybe rename this to "Persönleche Team" since it is translated, but maybe this also just causes confusion, @mtnstar what would you say about this?
If it should stay in english maybe type it capitatlized, because it's swissgerman so maybe something like "Personal Team"

config/locales/de.yml
```diff
@@ -81,6 +81,8 @@ de:
       weak_password: "Um die Sicherheit zu verbessern, sollten Sie ein komplexeres Passwort wählen"
     teams:
       not_existing: "Es existiert kein Team mit der Id %{id}"
+    folders:
+      duplicated_inbox: 'Es kann nur ein Inbox Ordner im personal Team geben'
```
njaeggi, line 85: Same with this as in the swissgerman translation, maybe change it to german and else use "Personal Team"

config/locales/en.yml
```diff
@@ -81,6 +81,8 @@ en:
       weak_password: "To improve security you should change your login password to a more complex one"
     teams:
       not_existing: 'Es existiert kein Team mit der Id %{id}'
```
njaeggi, line 83: This isn't your change but there's a german translation in english here? :thinking:

config/locales/fr.yml
```diff
@@ -81,6 +81,8 @@ fr:
       weak_password: "Pour améliorer la sécurité vous devez déterminer un mot de passe utilisateur plus complexe."
     teams:
       not_existing: 'Il n''y a pas d''équipe avec l''id %{id}'
+    folders:
+      duplicated_inbox: 'Il ne peut y avoir que un inbox folder dans le personal team'
```
njaeggi, line 85: Also here, maybe  translate personal team with "équipe personnelle"

frontend/app/models/folder.js
```diff
@@ -3,6 +3,11 @@ import Model, { attr, hasMany, belongsTo } from "@ember-data/model";
 export default class Folder extends Model {
   @attr("string") name;
   @attr("string") description;
+  @attr("number") unreadTransferredFiles;
```
njaeggi, line 6: maybe use a generalized variable name, because this is also used by credentials later, not only files, maybe use "unreadTransfersCount" or something similar to that

frontend/app/models/team.js
```diff
@@ -7,10 +7,24 @@ export default class Team extends Model {
...
   @attr("boolean") deletable;
+  @attr("number") unread_count;
   @hasMany("folder") folders;
   @hasMany("teammember") teammembers;
 
   get isPersonalTeam() {
     return this.type === "Team::Personal";
   }
+
+  get unreadTransferredFilesInFolders() {
```
njaeggi, line 18: also here maybe use a method name like "unreadTransfersInFolder" to generalize it for credentials share, then we don't have to rename anything when implementing the credential share feature :wink: 

Also is this supposed to be plural, because you only check inbox folder right?

frontend/translations/de.yml
```diff
@@ -90,6 +93,21 @@ de:
...
+    share:
+      title: Datei teilen
+
+  encryptable/transferring/file:
+    new:
+      title: "Datei teilen"
+    receiver:
+      name: Wähle einen Empfänger aus
+    show:
+      sender_name: Sender Name
```
njaeggi, line 105: Maybe use "Name vom Sender"

njaeggi (changes requested):

config/locales/ch_be.yml
njaeggi: @mtnstar said that you can use what you think works better either "Persönleche Team" or "Personal Team"

app/policies/encryptable_policy.rb
lkleisa: No it is the method which checks if the encryptable is transferring when sharing a new one, because when you share an encryptable, then the sender can't be teammember if the receiver's personal team.

mtnstar:
image

now the team texts are aligned a bit strange ?! 
did this come with your changes?
text should be left aligned I think

mtnstar:
image

the badge with unread count looks better now, but still a bit ugly :joy: 

maybe make it a bit smaller, bold text, and make sure it's centered vertically?

how does the user know to look inside the inbox for new messages? it seams not very obvious to me. we can't have a badge there too, can we?

app/controllers/api/encryptables_controller.rb
```diff
@@ -128,4 +142,68 @@ def permitted_attrs
...
+  ### Entries ###
+
+  def fetch_entries
+    return fetch_encryptable_files if credential_id.present?
+
+    user_encryptables
+  end
+
+  def render_entry(options = nil)
+    return send_file(options) if encryptable_file? && action_name == 'show'
```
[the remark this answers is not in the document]
mtnstar: so if the user is downloading a file, the content type is still json? but this is probably only possible if we use js to download the file. can't we just use a plain anchor link for downloading? so we do have to scenarios here:

1. we fetch the encryptable file as json for getting it's meta information and display it on encryptable#show in the frontend. 
2. we're sending it to the user for downloading

app/models/folder.rb
```diff
@@ -10,26 +10,29 @@
...
 
-#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
-#  Cryptopus and licensed under the Affero General Public License version 3 or later.
-#  See the COPYING file at the top-level directory or at
-#  https://github.com/puzzle/cryptopus.
-
 class Folder < ApplicationRecord
   belongs_to :team
-  has_many :encryptables, -> { order :name }, dependent: :destroy
+  has_many :encryptables, dependent: :destroy
```
mtnstar, line 18: can we just remove default order here? this impacts all other teams too ... order by name seams pretty handy to me? you might add another order condition for the inbox folder to prioritize it?

` -> { order [:personal_inbox, :name] }`

app/models/team/personal.rb
```diff
@@ -27,4 +27,12 @@ def create(owner)
       team
     end
   end
+
+  def unread_count_transferred_files
+    return nil if folders.where(name: 'inbox').empty?
+
+    # rubocop:disable Metrics/LineLength
+    folders.where(name: 'inbox').first.encryptables.all.where.not(encrypted_transfer_password: nil).count
```
mtnstar, line 35: we do have a boolean on File to get inbox folder, don't we?

don't ignore rubocop here ... use line breaks :wink: 

maybe add getter on personal team for inbox folder. 

also consider a getter on folder to get unread count. then you just might call it here.

app/serializers/folder_serializer.rb
```diff
@@ -18,8 +18,20 @@
...
+  has_many :encryptables, serializer: EncryptableMinimalSerializer do
+    if object.personal_inbox?
+      object.encryptables.order('created_at DESC')
+    else
+      object.encryptables.order(:name)
+    end
+  end
+
+  def unread_transferred_count
+    # rubocop:disable Metrics/LineLength
```
mtnstar, line 32: rubocop is telling you that this line is to long, because it's tooooooo long :joy: 

don't just add ignore statements here for this rule. rubocop should only be ignored if really necessary

frontend/app/models/team.js
```diff
@@ -7,10 +7,24 @@ export default class Team extends Model {
...
 
   get isPersonalTeam() {
     return this.type === "Team::Personal";
   }
+
+  get unreadTransfersInFolder() {
+    if (!this.isPersonalTeam) return undefined;
+    if (this.unread_count === null) return undefined;
+
+    let folder = this.folders.filter((folder) => folder.name === "inbox")[0];
```
mtnstar, line 22: const instead of let?

mtnstar (changes requested):
almost there ! looking pretty good ... but some important stuff to check!

go go go :rocket:

app/models/folder.rb
kcinay055679: Lias told me it's unnecessary at this point since it gets overwritten anyway in Folder Serializer

app/models/folder.rb
```diff
@@ -10,26 +10,32 @@
...
   end
+
+  def only_one_personal_inbox
+    if personal_inbox && team.folders.where(personal_inbox: true).where.not(id: id).exists?
+      errors.add(:personal_inbox, I18n.t('flashes.folders.duplicated_inbox'))
+    end
+  end
+
+  def unread_count_transferred_files
+    encryptables.all.where.not(encrypted_transfer_password: nil).count
```
mtnstar, line 39:
```suggestion
    encryptables.where.not(encrypted_transfer_password: nil).count
```

app/models/folder.rb
```diff
@@ -10,26 +10,32 @@
...
+  has_many :encryptables, dependent: :destroy
 
   attr_readonly :team_id
 
   validates :name, presence: true
   validates :name, uniqueness: { scope: :team }
   validates :name, length: { maximum: 70 }
   validates :description, length: { maximum: 300 }
-
+  validate :only_one_personal_inbox
```
mtnstar, line 26: may we use default arel functionality for this? https://apidock.com/rails/ActiveRecord/Validations/ClassMethods/validates_uniqueness_of

app/models/team/personal.rb
```diff
@@ -27,4 +27,15 @@ def create(owner)
       team
     end
   end
+
+  def unread_count_transferred_files
+    return nil if inbox_folder.nil?
+
+    inbox_folder.unread_count_transferred_files
```
mtnstar, line 34:
```suggestion
def unread_count_transferred_files
    inbox_folder&.unread_count_transferred_files
end
```

app/models/team/personal.rb
```diff
@@ -27,4 +27,15 @@ def create(owner)
...
+  end
+
+  def inbox_folder
+    folders.where(name: 'inbox').first
```
mtnstar, line 38:
```suggestion
    folders.find_by(personal_inbox: true)
```

mtnstar:
image

now the badge is perfectly vertically-aligned and styled for the inbox folder. :partying_face: 

can't we have the same for the badge at the personal team?

does it have to red the badge? it was red, than green, now it's red again? it's the only red item in the whole UI. maybe use a more decent color?

I know ... gränni :joy:

spec/unit/serializers/team_list_serializer_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+#  Copyright (c) 2008-2017, Puzzle ITC GmbH. This file is part of
```
mtnstar, line 3: 2017 ? :wink: 

I think it's ok to remove all license headers for now ...

mtnstar (changes requested):
:partying_face:  getting there !

check some small comments

don't forget to add Changelog entry, and update Version to 5.0

app/models/folder.rb
lkleisa: Thanks for the info. 
Changed it to validates :personal_inbox, because rubocop is gränning:
image

lkleisa: I let the scope blank, because personal_inbox is a boolean attribute on folder model and this boolean only can be true on one folder

lkleisa:
Please use squash commits when merging this pr

PR #689, opened 2023-05-08
Feature-685: Backend upgrade

lkleisa:


spec/system/nav_bar_system_spec.rb
```diff
@@ -7,10 +7,9 @@
 
   it 'shows admin button as admin' do
     login_as_user(:admin)
-    visit('/')
 
     within(page.find('pzsh-menu')) do
-      expect(page).to have_text('Admin').twice
+      expect(page.body.downcase).to have_text('admin', count: 4)
```
njaeggi, line 12: Why did this change from twice to count 4?

spec/system/nav_bar_system_spec.rb
```diff
@@ -20,7 +19,7 @@
     login_as_user(:tux)
 
     within(page.find('pzsh-menu')) do
-      expect(page).to have_text('Admin')
+      expect(page.body.downcase).to have_text('admin')
```
njaeggi, line 22: Isn't it possible anymore to check the old way (without the downcase and check for "Admin" and not "admin")

njaeggi (changes requested):
Looks really good to me, I just had two questions about some small changes

njaeggi:
I just teasted the application a bit and noticed that when you create a new credential, the password field in the form isn't hidden. I tried on Integration and Production and it is hidden there. This most likely isn't from this Pull Request tho and is fixed anyways by @njaeggi with the custom attr feature, since that changes the whole encryptable form anyways. Besides that it looked really good :rocket:

spec/system/nav_bar_system_spec.rb
Robin481: Same question :grin: and why the downcase?

spec/fabricators/encryptables/credentials_fabricator.rb
```diff
@@ -10,7 +10,7 @@
   name { Faker::Team.creature }
   cleartext_username { Faker::Internet.user_name }
   cleartext_password { Faker::Internet.password }
-  before_save do |account, attrs|
+  before_create do |account, attrs|
```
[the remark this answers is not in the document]
lkleisa: No, since the upgrade it's recommended to change from before_save to before_create and this do the same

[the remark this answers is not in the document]
Robin481: I am sure this is correct but if you already take some of your time to answer me why not link the recommendation as well? :wink:

config/application.rb
```diff
@@ -67,6 +67,16 @@ class Application < Rails::Application
     # Configure the default encoding used in templates for Ruby 1.9.
     config.encoding = "utf-8"
```
mtnstar, line 69: comment why this is needed now? introduced  by new rails version?

.github/workflows/build.yml
```diff
@@ -42,10 +42,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Set up Ruby 3.0
+      - name: Set up Ruby 3.2.2
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.0'
+          ruby-version: '3.2.2'
```
mtnstar, line 48: should be pretty stable if we just define version `3.2` here ... that would make sure we're getting minor version upgrades automatically

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.0
+FROM ruby:3.2.2
```
mtnstar, line 1: same ... verision `3.2` ... or is there a special reason why we should use minor version here?

mtnstar (changes requested):
nice ... very important upgrade !

some small things to check ... is build passing?

spec/fabricators/encryptables/credentials_fabricator.rb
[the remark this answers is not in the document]
lkleisa: I didn't find anything in a documentation, but here is my terminal output when I run my tests with the old version: 
image

config/application.rb
lkleisa: This comes from new rails version
https://stackoverflow.com/questions/72970170/upgrading-to-rails-6-1-6-1-causes-psychdisallowedclass-tried-to-load-unspecif
https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017
@Robin481 told me how he solved it in his projects

Dockerfile
lkleisa: No, no reason why defining special minor version here, I changed it.

config/application.rb
mtnstar: good, I added those links as comment just above this config block

Issue #634, opened 2022-06-03
ENCRYPTABLES: CUSTOM ATTRS: Changes to frontend #2


luca-bruegger:
In order to get the frontend working with the backend adjustments, some changes in the UI are needed.

![encryptables_create.png](https://images.zenhubusercontent.com/5faa4de0399793e6e36980fc/0fb3c9a2-1eb5-4ee2-a55a-32304daea039)

# ToDo

- [x] Change form to new mockup
- [x] add locks for encrypted fields
- [x] let fields be removed and added back
- [x] adjust show encryptable
- [x] adjust encryptable row
- [x] imrpove copy blank value handling
- [x] add new field component
- [x] adjust format-cleartext helper
- [x] add translations
- [x] adjust frontend model
- [x] add new model attribute used attrs for row
- [x] fetch encryptable for edit in row
- [x] change styling of row component
- [x] load form field dynamically
- [ ] Auf Integrations extensiv testen
- [ ] @mtnstar mit Proddaten testen bevor prod release

PR #690, opened 2023-05-08
Feature/633/custom attrs

njaeggi:
implements #634


lkleisa (dismissed):
Huge pull request, very nice implementation! 

Some UI things: 
- When I want to upload a file to an encryptable, the button is now much longer than before:
image
- The validation for the custom Label is not working:
image
- There is a missing validation when the name is too long (no matter if pin, email, token):
image

frontend/app/models/encryptable-credential.js
```diff
@@ -5,13 +5,58 @@ import { isNone } from "@ember/utils";
 export default class EncryptableCredential extends Encryptable {
   @attr("string") cleartextUsername;
   @attr("string") cleartextPassword;
+  @attr("string") cleartextToken;
+  @attr("string") cleartextPin;
+  @attr("string") cleartextEmail;
+  @attr({ defaultValue: undefined }) cleartextCustomAttr;
```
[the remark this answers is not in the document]
njaeggi: We don't really have a data type here, since cleartextCustomAttr is a nested attribute that consists of the two values label and value, we would have a data type defined here, when we add a nested model as a file

  @hasOne("custuom-attr") cleartextCustomAttr;
  
  But in our case we don't really need a nested model, since value and label are just two string fields, which can be defined with the help of methods. Because of that the datatype of cleartextCustomAttr is in our case just a json formatted nested attribute

frontend/app/components/encryptable/form.js
```diff
@@ -98,6 +140,53 @@ export default class Form extends BaseFormComponent {
     this.changeset.folder = folder;
   }
 
+  @action
+  hideField(value) {
```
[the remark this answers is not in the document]
njaeggi: This is really rational, thank you! :rocket:

spec/unit/utils/encryptable_move_handler_spec.rb
```diff
@@ -90,7 +90,7 @@
 
 
     decrypted = credential.decrypt(new_folder.team.decrypt_team_password(bob, private_key))
-    expect(decrypted).to eq('password')
+    expect(decrypted).to eq({:label=>"testcustomattrlabel", :value=>"testcustomattrvalue"})
```
njaeggi, line 93: Custom Attribute is the last value that is decrypted, before the custom attr feature it was the password, we now just check for the encrypted value of custom attr instead of password

Gemfile.lock
```diff
@@ -365,6 +363,7 @@ GEM
     zeitwerk (2.5.4)
 
 PLATFORMS
+  ruby
```
Robin481, line 366: Why do we need this now?

app/controllers/api/encryptables_controller.rb
```diff
@@ -115,6 +115,7 @@ def model_serializer
     "#{model_class.name}Serializer".constantize
   end
 
+  # rubocop:disable Metrics/MethodLength
```
Robin481, line 118: you could outsource the whole
```
[:cleartext_username, :cleartext_password, :cleartext_token,
:cleartext_pin, :cleartext_email, :folder_id,
{ cleartext_custom_attr: [:label, :value] }]
```
into  a method and just do `permitted_attrs + credential_attrs`

But not sure if that is much more readable but would probably remove the need for the `rubocop:disable` calls

app/models/encryptable/credentials.rb
```diff
@@ -1,25 +1,47 @@
...
            foreign_key: :credential_id,
            dependent: :destroy
 
+  def validate_at_least_one_attribute_set
+    attributes = [
+      cleartext_password, cleartext_username,
+      cleartext_token, cleartext_pin,
+      cleartext_email, cleartext_custom_attr
+    ]
+    unless attributes.any?(&:present?)
```
Robin481, line 18: isn't `attributes.any?` enough?

app/models/encrypted_data.rb
```diff
@@ -30,6 +38,10 @@ def data_hash(iv, data)
     { iv: iv, data: data }
   end
 
+  def data_hash_label(iv, label, data)
```
Robin481, line 41: no need for a seperate method, just do something like that: 
```
def data_hash(iv, data, label)
  hash = { iv: iv, data: data }
  hash[:label] = label if label
  hash
end
```

app/models/encrypted_data.rb
```diff
@@ -13,11 +13,19 @@ def [](key)
...
+      data_hash_label(iv, value[:label], decode(value[:data]))
+    else
+      data_hash(iv, decode(value[:data]))
+    end
   end
 
-  def []=(key, data:, iv:)
-    @data[key] = data_hash(encode(iv), encode(data))
+  def []=(key, data:, iv:, label: nil)
+    @data[key] = if label&.present?
```
Robin481, line 24: `if label` would probably work just as well, however this won't be needed anymore anyway when you look at the next comment :)

app/serializers/encryptable_minimal_serializer.rb
```diff
@@ -1,5 +1,19 @@
 # frozen_string_literal: true
 
 class EncryptableMinimalSerializer < ApplicationSerializer
-  attributes :id, :name, :description
+  attributes :id, :name, :description, :used_attrs
```
Robin481, line 4: what does that do and where is it used?

app/serializers/encryptable_minimal_serializer.rb
```diff
@@ -1,5 +1,19 @@
...
+
+  # rubocop:disable Metrics/AbcSize
+  def used_attrs
```
Robin481, line 7: Simplify with something like that (don't just copy paste pls check) : 
```
def used_attrs
  keys = [:password, :username, :token, :pin, :email, :custom_attr]
  data = object.encrypted_data.instance_variable_get(:@data)

  keys.each_with_object({}) do |key, used_attrs|
    used_attrs[key] = !data[key].nil?
  end
end
```

frontend/app/components/encryptable/form.js
```diff
@@ -7,6 +7,7 @@ import { tracked } from "@glimmer/tracking";
 import BaseFormComponent from "../base-form-component";
 import { isPresent } from "@ember/utils";
 import { isEmpty } from "@ember/utils";
+import { A } from "@ember/array";
```
Robin481, line 10: A

frontend/app/components/encryptable/form.js
```diff
@@ -16,10 +17,51 @@ export default class Form extends BaseFormComponent {
...
+  isEmailFieldActive = this.record.cleartextEmail;
+
+  //set display field to true when attribute is set
+  @tracked
+  isCustomAttrFieldActive = this.record.cleartextCustomAttr;
+
+  //create array proxy that ember checks the changes in each loop for dropdown, ember's rendering engine
+  // doesn't get notified by using a normal array
+  @tracked
+  items = A(
```
Robin481, line 51: A

(is it really just `A`? ... seems way too funny to me :joy: )

frontend/app/components/encryptable/form.js
```diff
@@ -98,6 +140,53 @@ export default class Form extends BaseFormComponent {
     this.changeset.folder = folder;
   }
 
+  @action
+  removeField(value) {
```
Robin481, line 144: huh do you really need to do all this by hand?
seems overly complicated.

frontend/app/components/encryptable/form.js
```diff
@@ -98,6 +140,53 @@ export default class Form extends BaseFormComponent {
...
+      this.changeset.cleartextEmail = null;
+    } else if (value === "customAttr") {
+      this.isCustomAttrFieldActive = false;
+      this.items.addObject(value);
+      this.changeset.cleartextCustomAttr = null;
+    }
+  }
+
+  @action
+  addField() {
```
Robin481, line 173: same here, seems very complicated for what is a simple form

frontend/app/components/encryptable/form.js
```diff
@@ -134,7 +223,10 @@ export default class Form extends BaseFormComponent {
   }
 
   handleSubmitError(response) {
-    this.hasErrors = response.errors.length > 0;
+    if (response.errors.length > 0) {
+      //get error message to show correct error in form
+      this.Errors = response.errors[0].detail;
```
Robin481, line 228: This will only give you the first error though.

frontend/app/components/encryptable/row.js
```diff
@@ -15,30 +15,44 @@ export default class RowComponent extends Component {
...
+  @tracked
+  isCustomAttrVisible = false;
 
   @tracked
   isShown = false;
 
-  @action
-  copyPassword() {
-    this.fetchAndCopyToClipboard("password");
+  //get amount of currently set attributes on encryptable, because all encryptables with more than two set attributes
```
Robin481, line 46: space  and i don't really understand the comment... reword :)

frontend/app/components/encryptable/row.js
```diff
@@ -109,22 +135,26 @@ export default class RowComponent extends Component {
   }
 
   @action
-  showPassword() {
+  showAttribute(attribute) {
     this.fetchEncryptable();
-    this.isPasswordVisible = true;
+    this[
+      `is${attribute.charAt(0).toUpperCase()}${attribute.slice(1)}Visible`
```
Robin481, line 141: what does this do?

frontend/app/components/encryptable/row.js
```diff
@@ -109,22 +135,26 @@ export default class RowComponent extends Component {
...
-      let passedTime = now - this.passwordHideCountdownTime;
+      let passedTime = now - this.hideCountdownTime;
 
       let passedTimeInSeconds = Math.floor(passedTime / 1000);
 
       if (passedTimeInSeconds >= this.HIDE_TIME) {
-        this.isPasswordVisible = false;
-        clearInterval(this.passwordHideTimerInterval);
+        this[
+          `is${attribute.charAt(0).toUpperCase()}${attribute.slice(1)}Visible`
```
Robin481, line 155: this hides the passwort after some time... seems overly complicated though... and should probably be extracted into a method if used several times

frontend/app/components/encryptable/show.js
```diff
@@ -35,8 +47,18 @@ export default class ShowComponent extends Component {
   }
 
   @action
-  showPassword() {
-    this.isPasswordVisible = true;
+  showValue(value) {
```
Robin481, line 50: yea this could be prettier as well...
Maybe something like this
```
showValue(value) {
  const visibilityMap = {
    "password": "isPasswordVisible",
    "pin": "isPinVisible",
    "token": "isTokenVisible",
    "email": "isEmailVisible",
    "customAttr": "isCustomAttrVisible",
  };

  if (visibilityMap.hasOwnProperty(value)) {
    this[visibilityMap[value]] = true;
  }
}
```
this at least gets rid of all the if else jungle.
Nevertheless it isn't much more concise, maybe a different value could be passed?

frontend/app/models/encryptable-credential.js
```diff
@@ -5,13 +5,58 @@ import { isNone } from "@ember/utils";
...
+  get isEmailBlank() {
+    return this.isFullyLoaded && isNone(this.cleartextEmail);
+  }
+
+  get isCustomAttrBlank() {
+    //make it null save because custom attr is nested
+    return (
+      this.isFullyLoaded &&
+      isNone(
+        (this.cleartextCustomAttr == null ? "" : this.cleartextCustomAttr).value
```
Robin481, line 58: why is this needed?

frontend/app/styles/app.scss
```diff
@@ -192,6 +193,21 @@ li a {
...
+
+@media (max-width: 1062px) {
+  .encryptable-row-attribute,
+  .encryptable-row-copy-button {
+    display: none !important;
+  }
+}
+
+.file-upload-button {
+  width: 17% !important;
```
Robin481, line 208: is important really needed?

frontend/app/styles/app.scss
```diff
@@ -192,6 +193,21 @@ li a {
...
 }
 
+.encryptable-row-icons {
+  margin-left: auto;
+}
+
+@media (max-width: 1062px) {
+  .encryptable-row-attribute,
+  .encryptable-row-copy-button {
+    display: none !important;
```
Robin481, line 203: important really needed?
and why do we hide this button if the screen is small?

frontend/app/styles/encryptable-form.scss
```diff
@@ -0,0 +1,60 @@
+.encryptable-form-name {
+  border-bottom-color: rgb(222, 226, 230);
+  border-bottom-style: solid;
+  border-bottom-width: 1px;
+}
+
+.add-field {
```
Robin481, line 7: please use more bootstrap wherever possible... the less custom classes the better. give yourself quite a bit of time to work through your css and replace it with bootstrap wherever possible

frontend/translations/fr.yml
```diff
@@ -19,6 +19,13 @@ fr:
   tag: Tag
   username: Nom d'utilisateur
   password: Mot de passe
+  pin: Épingle
```
Robin481, line 22: leave it at Ping

Robin481 (changes requested):
Nice changes but a few improvements here and there :3rd_place_medal:

Robin481 (changes requested):
Nice changes but a few improvements here and there :3rd_place_medal:

Gemfile.lock
njaeggi: this shouldn't be there, removed it 3 times already and added it back again :disappointed_relieved:

app/controllers/api/encryptables_controller.rb
njaeggi: I tried it already, works the same, but for me it was less readable, so I removed it and disabled rubocop for this. But I can still outsource it, if it's wanted

app/serializers/encryptable_minimal_serializer.rb
njaeggi: This is used in frontend in row component, to determine  what values are set in the encryptable, in the row component we only get the minimal data, and only when user reveals or copies the attribute, the encryptable is decrypted and sent again with the values. Because we now check how many attributes are used to display the correct amount of fields in the row, we need extra information from the backend.

This also can't just be a number of how many are used, because in that case we don't know what attributes are set and can't display them with the right values.

frontend/app/components/encryptable/form.js
njaeggi: B

frontend/app/components/encryptable/form.js
njaeggi: Seems like it is...

https://api.emberjs.com/ember/release/classes/ArrayProxy

frontend/app/components/encryptable/row.js
njaeggi: this[`is${capitalize(attribute)}Visible`] = true;

Basically the same as this, changed it to what I used in other Places

frontend/app/components/encryptable/form.js
njaeggi: Yes, I will implement multiple error handling, I did this in a rush and we were also just checking for one error before.

app/controllers/api/encryptables_controller.rb
Robin481: yea leave it as it is :v:

frontend/app/styles/app.scss
MarcEgliP: Class "ecryptable-row-icons" has been removed and reimplemented via Bootstrap. The media query stays because we have to hide the button otherwise the row wouldn't fit properly on the screen.

frontend/translations/fr.yml
MarcEgliP: Has been translated again.

frontend/app/styles/app.scss
MarcEgliP: Couldn't find the button inside the browser which uses this css class. Is this even needed?

frontend/app/styles/encryptable-form.scss
MarcEgliP: Bootstrap was implemented in encryptable/forms.hbs

frontend/app/styles/app.scss
MarcEgliP: Has been resolved.

frontend/app/components/encryptable/form.js
njaeggi: We have todo all this, but I generalized it to 3 lines instead of this if else construct :disappointed_relieved:

MarcEgliP (dismissed):

frontend/app/components/encryptable/form.js
Robin481: Good job!

Robin481 (changes requested):
I think the code is getting to a good place.
I'll start with a functionality review and we can then review the code again after these changes.  
 
1)
image
On load the custom field is not aligned with the rest of the fields.

2)
image
Not sure if this placeholder is intuitive to users. A quick fix would be to have it called "choose an additional field" and then grey out the text. Only make the actual options properly black. Furthermore the "Additional field" text shouldn't show up in the dropdown, doesn't make much sense.
I feel like there is a lot of usability improvement still waiting to happen. How about you don't even have the green button and just have the dropdown that says "choose an additional field" and then `onClick()` on one of the dropdown options it just adds said field. 

3)
image
If you press the green plus button in this state there should probably be a warning message like "choose one of the options in the dropdown". However, as mentioned above maybe just remove the button as it doesn't really add any value. (I know that it was in the mockup but we can be better than the mockup :wink: )

4)
image
Add an on-hover message to this lock that says "this field is encrypted" and then maybe also add a crossed-out-lock to the `custom attribute name` that `onHover` says "this field is not encrypted"

5)
image
The order in which the fields then are displayed is a bit confusing. I think `username` and `password` should always be the first 2 options as they are the default ones anyways. For the other ones it is fine if they are in no particular order, just not in the first or second spot. 

6) 
The dropdown has a few funky interactions. One example: go to a new encryptable, add a new custom attribute by selecting it from the dropdown and pressing the green plus button. Then instantly delete the new custom attribute and press the green button again. Even though the option "Additional field" is selected it will once more add the custom attribute.

Robin481:
image
Also add a little grey border around those 2 fields so you can see that they belong together.

frontend/app/validations/encryptable.js
```diff
@@ -1,12 +1,41 @@
...
+  ],
+  cleartextPin: [
+    validateExclusion({ list: [""] }),
+    validateLength({ max: 70 })
+  ],
+  cleartextEmail: [
+    validateExclusion({ list: [""] }),
+    validateLength({ max: 70 })
+  ],
+  label: [
```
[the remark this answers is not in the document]
njaeggi: This is correct, because we check for each others presence, changed the length validation between the two

njaeggi:
All UI changes have been implemented, besides the new svg for non encrypted fields, everything else is resolved.

MarcEgliP:
Was noch gemacht werden muss:
- Validierung bei Form funktioniert noch nicht zu 100% => Wenn ich neue Credentials erstellen will, und nachdem ich den Accountnamen definiert habe direkt auf "submit" klicke, entsteht ein Fehler und das Frontend schmiert ab.

Aufgetretene Fehler / Abgehandeltes:
- Wenn neue Credentials erstellt werden und man das Passwort eingibt, wird es nicht geblurred, dieser issue wird aber in einen anderen Ticket #704 behandelt.
- Das Icon (durchgestrichen) konnte eingefügt werden
- Alle Frontend Tests wurden gefixed
- Ein Border wird nur um das Custom Attributes Feld gemacht

Robin481:
Quick feedback for @njaeggi 

Bugs to fix
1) Password not hidden anymore
2) Alignment of custom field dropdown
3) Label doesn't get removed when all the fields are selected

Robin481 (approved):
Seems fine to me! :rocket: The UI is now quite nice if you ask me. :partying_face: 
However, let @mtnstar also review it one more time.

mtnstar:
image

styling? make sure all attribute boxes and copy controls look the same and are aligned. maybe give them more width?
or use two columns?

mtnstar:
same as above for the list view ... 

image

maybe we could align the value/copy boxes to the right?

mtnstar:
image

on encryptables#show ... password is shown in clear text on my local dev machine. 

image

should be with the secret font ... maybe I just missed to install something?

mtnstar:
image


is there a reason why there's a blank value inside this drop-down? can't we just pre-select any of the options and the user might change it?

frontend/app/components/encryptable/form.js
mtnstar: > A
> 
> (is it really just `A`? ... seems way too funny to me joy )

it's JavaScript :joy:

app/serializers/encryptable/credentials_serializer.rb
```diff
@@ -1,7 +1,18 @@
 # frozen_string_literal: true
 
 class Encryptable::CredentialsSerializer < EncryptableSerializer
-  attributes :cleartext_username, :cleartext_password, :created_at, :updated_at
+  attributes :cleartext_username, :cleartext_password, :cleartext_token, :cleartext_pin,
+             :cleartext_email, :cleartext_custom_attr, :created_at, :updated_at
+
+  def cleartext_custom_attr
```
mtnstar, line 7: can't we just have a separat attribute called `cleartext_custom_attr_label` for the label?
so we would not have to handle the cleartext value differently for the custom attr.
I don't like the idea of having a special structure here for the custom attr.

mtnstar (changes requested):
guys, I'm really looking forward to this feature! it's huge, well done so far and almost there!

some things to check:

1. design and maybe also usability must be improved
2. try to make your code more DRY, extract code to reusable fragments where possible
3. make sure `encrypted_data` is just a dumb data container and validate it's possible key/value pairs inside encryptable::credentials model

njaeggi:
> image
> 
> is there a reason why there's a blank value inside this drop-down? can't we just pre-select any of the options and the user might change it?

UX defined that it would be more user friendly and understandable for the user with a "placeholder" inside the dropdown, should we change it to a selected value?

frontend/app/templates/components/encryptable/form.hbs
```diff
@@ -1,107 +1,277 @@
...
+                              this.changeset.error.description.validation "invalid-input"}} vertical-resize"
+                            customError={{t
+                (validation-error-key this.changeset.error.description.validation)}} as |el|>
+                <el.control tabindex="7"></el.control>
+              </form.element>
+            </form.group>
+          </div>
+          <div class="encryptable-form-attributes">
+            <div class="row">
+              {{#if this.isUsernameFieldActive}}
```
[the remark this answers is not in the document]
njaeggi: When we extract the field into a separate components, there is a compiler error for all properties set on form.element, like @property, @customError or @label. This is the case because the form.element isn't in the same component as the opening and closing tag of BsForm.

The BsForm basically opens and closes the whole form and with it the possibility to use form.group and form.element. I also tried passing down the form into the new child component for the fields and use it like that, but since we can't use mustaches in element tag names this approach also doesn't work.

I saw other people having the same issues and they got the recommendation to leave the form in one solid component, and that it doesn't really make sense to split it into multiple child components.

Maybe you have a solution, or you think it is fine when we leave it as it is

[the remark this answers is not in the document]
njaeggi: A different approach to make the code more DRY and reduce the length of the file would be to remove the booleans for the fields completely and add a second array for all active items, so we can switch the currently active fields from an array called activeFields to an array called inactiveFields.

With that we could iterate over it in the component and load in the fields one by one.

Advantages to current form:
- more DRY
- less file length
- when adding a new field, it will always display at the end of the form, e.g. currently when only having password field active and adding the usename field, or whatever field that comes before password in out template, it adds the username before the password field and moves every following field in the form, which is kinda confusing. With the new way with the arrays the new element will be placed as the last element.

Disadvantages:
- Template could get more complex, since we still need if statements inside the each loop of the password field and custom attrs field.

For me this approach would be the best solution, since we make it more user friendly and the file will be much shorter. The complexity we add is basically already in the template right now, just implemented differently.

I suggested the approach to @Robin481 and he told me to ask you @mtnstar, what you think about this way of approaching this feedback.

MarcEgliP:
Was gemacht wurde:
- Delete Button auf Custom-Attr-Label Ebene entfernt
- Rows der encryptable wieder kongruent zu den Rows auf Produktion gemacht
- Responsivness wiederhergestellt

Was noch gemacht werden muss:
- Wenn ich mehrere Attribute wie Pin, Email und Token hinzufüge, speichere und dann per edit button versuche die Attribute wieder zu löschen, sendet der Save Button kein PostRequest mehr, das Frontend reagiert auch nicht mehr => Muss gelöst werden

njaeggi:
Ich kann den Bug den Marc beschrieben hat auch nicht reproduzieren

app/models/encryptable.rb
```diff
@@ -91,6 +98,10 @@ def decrypt_attr(attr, team_password)
                         Crypto::Symmetric::Aes256.decrypt(encrypted_value, team_password)
                       end
 
+    if attr == :custom_attr
+      instance_variable_set("@cleartext_#{attr}_label", encrypted_data[attr].try(:[], :label))
```
mtnstar, line 102:
```suggestion
      @cleartext_custom_attr_label, encrypted_data[:custom_attr][:label]))
```

app/models/encryptable/credentials.rb
```diff
@@ -11,15 +12,25 @@ class Encryptable::Credentials < Encryptable
   validates :name, length: { maximum: 70 }
   validates :name, uniqueness: { scope: :folder }
   validates :folder_id, presence: true
+  # validates :cleartext_custom_attr, presence: { if: :cleartext_custom_attr_label }
```
mtnstar, line 15: not required anymore? commented?

mtnstar: as we discussed already ... if no label set, just assume a default like `custom`. so you don't have to validate it

app/serializers/encryptable_minimal_serializer.rb
```diff
@@ -1,9 +1,18 @@
 # frozen_string_literal: true
 
 class EncryptableMinimalSerializer < ApplicationSerializer
-  attributes :id, :name, :description, :created_at, :sender_name
+  attributes :id, :name, :description, :created_at, :sender_name, :used_attrs
```
mtnstar, line 4:
```suggestion
  attributes :id, :name, :description, :created_at, :sender_name, :used_encrypted_data_attrs
```
I know, longer term like this, but more clear and specific

app/serializers/encryptable_minimal_serializer.rb
```diff
@@ -1,9 +1,18 @@
...
 
   def sender_name
     object.sender&.label
   end
+
+  def used_attrs
+    keys = [:password, :username, :token, :pin, :email, :custom_attr]
+    data = object.encrypted_data.instance_variable_get(:@data)
+
+    keys.index_with do |key|
```
mtnstar, line 14: why not #collect?

app/serializers/encryptable_minimal_serializer.rb
```diff
@@ -1,9 +1,18 @@
...
-  attributes :id, :name, :description, :created_at, :sender_name
+  attributes :id, :name, :description, :created_at, :sender_name, :used_attrs
 
   def sender_name
     object.sender&.label
   end
+
+  def used_attrs
+    keys = [:password, :username, :token, :pin, :email, :custom_attr]
+    data = object.encrypted_data.instance_variable_get(:@data)
```
mtnstar, line 12: move this whole logic to encryptable base model ... so if you have to access an instance variable from outside, something is probably not right :wink:

frontend/app/components/encryptable/attribute-field.js
```diff
@@ -0,0 +1,98 @@
...
+      this.startHideAttributeTimer();
+    }
+
+    this.isAttributeVisible = true;
+  }
+
+  startHideAttributeTimer() {
+    this.hideCountdownTime = new Date().getTime();
+
+    this.hideTimerInterval = setInterval(() => {
```
mtnstar, line 39: does this timer work on your machine?
once i revealed a secret, it's just being displayed forever

frontend/app/components/encryptable/attribute-field.js
```diff
@@ -0,0 +1,98 @@
+import Component from "@glimmer/component";
```
mtnstar, line 1: very nice this new component ! :partying_face:

frontend/app/components/encryptable/row.js
```diff
@@ -38,62 +22,11 @@ export default class RowComponent extends Component {
...
-  encryptableFullyLoaded(encryptable) {
-    if (encryptable.isPasswordBlank) {
-      this.isPasswordVisible = true;
-    }
-    if (encryptable.isUsernameBlank) {
-      this.isUsernameVisible = true;
-    }
+  // get set attribute amount to hide attribute fields in encryptable row when encryptable has more than two attributes
+  get getAttributesAmount() {
+    return Object.values(this.args?.encryptable?.usedAttrs ?? {}).filter(
```
mtnstar, line 27: we can not just call length on `this.args?.encryptable?.usedAttrs` ?

spec/controllers/api/encryptables_controller_spec.rb
```diff
@@ -10,7 +10,9 @@
   let(:api_user) { bob.api_users.create }
   let(:private_key) { bob.decrypt_private_key('password') }
   let(:nested_models) { ['folder'] }
-  let(:attributes) { %w[name cleartext_password cleartext_username] }
+  # rubocop:disable Metrics/LineLength
```
mtnstar, line 13: do not add an exception here ... rubocop is gränning because the line is really toooooo long :wink: 

use do end block and put arguments on mutlipe lines

mtnstar (changes requested):
wow, almost 2200 lines of new code ... definitely one of the biggest PR ever !

getting there ... some small things to check ... :rocket:

app/models/encryptable/credentials.rb
njaeggi: We validate it in frontend right now, should I also remove it?

njaeggi:
Was wurde gemacht:
- Verbesserungen an Frontend
- UsedAttrs anders gelöst
- Bug mit Validierung von label wurde behoben
- Bootstrap input groups für alles Felder eingebaut und Styling verändert
- Intensives Testing

Noch offen:
- n+1 query lösen
- deprecation warning in browserkonsole untersuchen

njaeggi:
vermutetes n+1 query war kein n+1 query sondern ein query welches immer alles Encryptable::File geladen hat, dies war bereits nach der file_sharing PR der Fall. Gemachte Änderung um diesen Bug zu fixen


    Encryptable::File.where(credential_id: user_encryptables)
    
    zu
    
    user_encryptables.where(type: Encryptable::File.sti_name).where(credential_id: credential_id)

njaeggi:
Deprecation warning in console should be fixed with the next ember upgrade

app/models/encryptable.rb
```diff
@@ -80,8 +84,14 @@ def encrypt_attr(attr, team_password)
                         Crypto::Symmetric::Aes256.encrypt(cleartext_value, team_password)
                       end
     # rubocop:enable all
+    build_encrypted_data(attr, encrypted_value)
+  end
 
-    encrypted_data.[]=(attr, **{ data: encrypted_value, iv: nil })
+  def build_encrypted_data(attr, encrypted_value)
+    attr_label = send(:cleartext_custom_attr_label) if attr == :custom_attr
```
mtnstar, line 91:
```suggestion
    attr_label = cleartext_custom_attr_label if attr == :custom_attr
```

mtnstar (approved):
very nice :partying_face: 

one small thing to check, then rebase && merge :rocket:

app/models/encryptable.rb
njaeggi: oops

PR #691, opened 2023-05-09
Automatic wake up

kcinay055679:


config/auth.yml
```diff
@@ -1,18 +1,18 @@
-provider: db
+provider: openid-connect
```
mtnstar, line 1: leave db as default !

public/wakeUp.js
```diff
@@ -0,0 +1,15 @@
+window.addEventListener("focus", () => {
```
mtnstar, line 1: wakeup.js or wake_up.js

usually, we do have app/javascript for js files in rails applications.

but it's a special case here since we usually use ember for frontend stuff, but not for the login page. 

maybe add this to a sub folder inside public? isn't it a bit messy like this?

public/wakeUp.js
```diff
@@ -0,0 +1,15 @@
...
+    redirect()
+});
+document.addEventListener("DOMContentLoaded", () => {
+    document.getElementById("submit").addEventListener("click", () => {
+       redirect()
+    })
+});
+
+function redirect() {
+    let path = sessionStorage.getItem("previousPath")
```
mtnstar, line 11: use `const` instead of let?

maybe clear value after redirect?

app/views/session/_oidc.html.haml
```diff
@@ -1,6 +1,7 @@
+%script{src: "/wakeUp.js"}
 .login-form
   = image_tag "/assets/images/cryptopussy_oidc_sleep.svg", class: 'logo'
   .form-group
     .alert.alert-login-box.alert-info
       = t('.sleep')
-    = link_to(t('.submit'), Rails.application.routes.url_helpers.root_path, class: 'btn btn-primary')
```
mtnstar, line 6: maybe just leave path if js is somehow disabled? maybe path is not required at all?!

with path, you might need prevent default on submit button.

mtnstar (changes requested):
gooood ! ... but .. PR is still draft ?!

some things to check

Robin481 (approved):
Seems fine by me, we'll have to test on INT as well if it actually works with Keycloak and such.

PR #692, opened 2023-05-09
include private field in serializer

MarcEgliP:
Was gemacht wurde:

- Private field von Teams wird nun auch nach vorne gereicht
- Private teams werden nun mit "encrypted_small_white.svg" Icon dargestellt


MarcEgliP:
Pipeline is failing due to unknown reasons. Changes which you can see in Github Tab are just one line inside the Teamlistserializer which shouldn't affect any of the failing tests. This feature is otherwise complete.

PR #703, opened 2023-05-22
Improve password generator

kcinay055679:
…password generation


kcinay055679:
Functional everything is fine.

TODO:
- [ ] Design
- [ ] Tests

frontend/app/components/encryptable/form.js
```diff
@@ -151,4 +160,8 @@ export default class Form extends BaseFormComponent {
       AccountValidations
     );
   }
+
+  get invertedWithSymbols() {
```
Robin481, line 164: Is this really needed or can you just do it at another point?
If yes why not name it `notWithSymbols`

frontend/app/styles/app.scss
```diff
@@ -465,3 +465,7 @@ h5, h6 {
   align-content: center;
   text-align: center;
 }
+
+#cleartext-password{
+  margin-bottom: 0.5em;
```
Robin481, line 470: Bootstrap?

frontend/app/templates/components/encryptable/form.hbs
```diff
@@ -94,6 +99,7 @@
                   <el.control tabindex="7"></el.control>
                 </form.element>
               </form.group>
+
```
Robin481, line 102: remove

Robin481 (changes requested):
Looking quite nice.
However, I don't think the interface makes a lot of sense at this point.
Currently I would argue 95% of the time a user doesn't create a password in Cryptopus but rather saves an already existing password.

Therefore, the use case of going to an encryptable which already has a password and randomly generating a new one that then automatically overwrites the old one doesn't really exist in my opinion and is rather confusing. 

Maybe a easy solution would be to not automatically overwrite the existing password but rather have a button that says "create new secure password" instead of always overwriting when using the slider or checking the symbol box.

Robin481 (changes requested):
Looking quite nice.
However the button takes up too mache space now.
What about putting it there and making it smaller? Think that would be the best of both worlds.
image

Robin481 (dismissed):
Looks very nice!
As discussed with @kcinay055679 a nice improvement would be to maybe not show the password generator when the field is already pre-filled. Work out what you think is a nice idea and let's discuss it :)

kcinay055679:
I added a accordion to the password generator witch is expanded on default and auto collapses after generating a new password. But the Accordeon generated by my code isnt the same as the one from the [docs](https://www.ember-bootstrap.com/components/accordion). I think the reason for that behavior is that we are currently using bootstrap 4.6.

Robin481:
@kcinay055679 
Good start.
Make the dropdown smaller, give it a bottom margin and add an arrow to the dropdown that actually indicates that this is a drodopwn, something like this 
image

kcinay055679:
The requested arrow is default functionality according to the ember docs, but in our project the arrow is simpy not there.
i think this is because we're using bootstrap 4

PR #707, opened 2023-05-31
Bootstrap update

kcinay055679:


TheWalkingLeek (changes requested):
image
The buttons should be further to the right no? Other than that, everything looks good :+1:

TheWalkingLeek (approved):

PR #708, opened 2023-05-31
Feature/688: Credentials sharing

lkleisa:


mtnstar:
image

sharing icon does not have same color as other buttons

mtnstar (changes requested):
another awesome feature I'm looking forward to it! also our members will appreciate it!

some small things to check :man_dancing:

app/controllers/api/encryptables_transfer_controller.rb
```diff
@@ -5,23 +5,35 @@ class Api::EncryptablesTransferController < ApiController
...
   end
 
+  def prepare_encryptable_credential
+    shared_encryptable = Encryptable::Credentials.find(params['encryptable_id'])
+
+    shared_encryptable.decrypt(decrypted_team_password(shared_encryptable.team))
+
+    @encryptable = shared_encryptable.dup
+
+    shared_encryptable.encrypt(decrypted_team_password(shared_encryptable.team))
```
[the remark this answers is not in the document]
lkleisa: I thought, when we decrypt the encryptable and duplicate it, we need to encrypt again for some security reason. But I removed the line

lkleisa:
# To do
- [ ] Fix rubocop
- [ ] Write test in encryptable_transfer_controller_spec, for sending to non existing user (raise error)

app/utils/encryptable_transfer.rb
```diff
@@ -28,6 +30,48 @@ def receive(encryptable, private_key, personal_team_password)
...
+        encryptable_name = "#{encryptable_name}(#{counter + 1})"
+      else
+        encryptable_name = "#{encryptable_name} (#{counter + 1})"
+      end
+    end
+
+    encryptable_name
+  end
+
+  def find_matching_inbox_names(new_encryptable_name, inbox_folder_names)
```
mtnstar, line 67:
```suggestion
  def find_matching_inbox_names(new_encryptable_name, inbox_folder_names)
```

better than matching, maybe existing_names?

app/utils/encryptable_transfer.rb
```diff
@@ -28,6 +30,48 @@ def receive(encryptable, private_key, personal_team_password)
...
+      # rubocop:enable Metrics/LineLength
+
+      encryptable_name = new_encryptable_name
+      counter += 1
+    end
+    encryptable_name
+  end
+
+  def generate_new_encryptable_name(encryptable_name, counter, matching_inbox_names)
+    clip_regex = / \(\d+\)/
```
mtnstar, line 52:
```suggestion
    INCREMENT_REGEX = / \(\d+\)/
```

also consider putting constants on top of the class ... like in java :wink:

app/utils/encryptable_transfer.rb
```diff
@@ -28,6 +30,48 @@ def receive(encryptable, private_key, personal_team_password)
...
+
+
+    matching_inbox_names.each do |inbox_encryptable_name|
```
mtnstar, line 55: maybe early return if there's no matches?

app/utils/encryptable_transfer.rb
```diff
@@ -28,6 +30,48 @@ def receive(encryptable, private_key, personal_team_password)
...
+      new_encryptable_name = generate_new_encryptable_name(encryptable_name, counter, matching_inbox_names)
+      # rubocop:enable Metrics/LineLength
+
+      encryptable_name = new_encryptable_name
+      counter += 1
+    end
+    encryptable_name
+  end
+
+  def generate_new_encryptable_name(encryptable_name, counter, matching_inbox_names)
```
mtnstar, line 51: take some time to refactor this method ... it's a bit messy and not so easy to understand.

- rename method to target_name or something similar?
- existing_names instead of matching_inbox_names?

mtnstar:
image

after I receive an encryptable, it would be awesome if I could move it to another folder or just edit it in any way. 

is it by intension those entries cannot be edited?

frontend/app/components/credential-transfer.js
```diff
@@ -0,0 +1,85 @@
+import { action } from "@ember/object";
```
mtnstar, line 1: awesome, very clean component ! :partying_face:

frontend/app/models/encryptable-transfer.js
```diff
@@ -0,0 +1,56 @@
...
+          description: this.description || "",
+          file: this.file,
+          receiver_id: receiverId
+        },
+        headers: {
+          "X-CSRF-Token": this.csrfToken
+        }
+      };
+
+      let promise = this.file.upload(targetUrl, opts);
```
mtnstar, line 45: const ?

frontend/app/models/encryptable-transfer.js
```diff
@@ -0,0 +1,56 @@
...
+      promise
+        .then((savedRecords) => {
+          let data = JSON.parse(savedRecords.body).data;
```
mtnstar, line 48: const ?

spec/controllers/api/encryptables_transfer_controller_spec.rb
```diff
@@ -75,4 +76,64 @@
...
+      login_as(:bob)
+
+      request_params = {
+        receiver_id: 139082421,
+        encryptable_id: credentials2.id
+      }
+
+      expect do
+        post :create, params: request_params, xhr: true
+      end.to raise_error(ActiveRecord::RecordNotFound, "User with id 139082421 does not exist")
```
mtnstar, line 91: make sure correct http code is returned, and also json error message so the frontend can handle it.

spec/controllers/api/encryptables_transfer_controller_spec.rb
```diff
@@ -75,4 +76,64 @@
...
+  it 'Does not send encryptable credentials which the sender does not have access to' do
+    login_as(:alice)
+    api_controller = ApiController.new
+
+    request_params = {
+      receiver_id: bob.id,
+      encryptable_id: credentials2.id
+    }
+
+    # There is no error raised when user don't have access
```
mtnstar, line 132: well, then raise an error in that case ! 
you probably need a special pundit policy for this case.
so `authorize entry` should fail in that case

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -56,5 +57,67 @@
...
+      expect(received_encryptable.name).to eq(encryptable_credentials1.name)
+
+      encryptable_transfer.transfer(encryptable_credentials1, receiver, sender)
+
+      received_encryptable = bob.inbox_folder.encryptables.last
+
+      expect(received_encryptable.name).to eq('Personal Mailbox (2)')
+    end
+
+    it 'Does not add (1) when name is not exact' do
```
mtnstar, line 102: what does it mean ... not exact? don't get it :see_no_evil: 

try to create less technical test cases, more from the user's perspective

mtnstar (changes requested):
niiiiice, staring to really love your new feature! some things to check ... keep going :rocket:

app/controllers/api/encryptables_controller.rb
```diff
@@ -110,7 +110,8 @@ def query_param
   end
 
   def encryptable_move_handler
-    EncryptableMoveHandler.new(entry, users_private_key, current_user)
+    entry.sender_id = nil
+    EncryptableMoveHandler.new(entry, session[:private_key], current_user)
```
lkleisa, line 114: Removed, because users_private_key needs `Authorization-Password` in request headers and we don't add them here. This method is working and is not bad, I think :smile:

spec/controllers/api/encryptables_transfer_controller_spec.rb
```diff
@@ -8,55 +8,83 @@
...
       login_as(:bob)
 
       file = fixture_file_upload('test_file.txt', 'text/plain')
+
       file_params = {
         content_type: 'text/plain',
         file: file,
         description: 'test',
-        receiver_id: alice.id
+        receiver_id: nil
```
mtnstar, line 69: use a non existing id here .... 42424424242 or something like that ... not nil

mtnstar (changes requested):
almost done :partying_face: 

just your logic to determine the target name really needs your attention ! :rocket:

app/controllers/api/encryptables_transfer_controller.rb
```diff
@@ -5,23 +5,35 @@ class Api::EncryptablesTransferController < ApiController
...
+
+    instance_variable_set(:"@#{ivar_name}", @encryptable)
```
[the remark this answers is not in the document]
lkleisa: No, you're right

lkleisa:
Start with feedback of mtnstar for the method which increases the name of the encryptable when the name already exists. 
# Todo 
- [ ] Fix rubocop
- [ ] Make the incasement working
- [ ] Smoke testing

frontend/app/templates/components/encryptable/row.hbs
```diff
@@ -35,51 +35,68 @@
...
+          </button>
+        </div>
+      {{/if}}
 
       <div class="col-md-auto d-none d-md-block">
-        <a class="mx-1" role="button" {{on "click" this.toggleEncryptableEdit}}>
-          <img class="icon-medium-button d-inline" src="/assets/images/edit.svg" alt="edit">
-          <BsTooltip @title={{t "encryptable/credentials.edit.title"}} @delayShow="1000" />
-        </a>
+        {{#unless this.isTransferredCredentials }}
```
lkleisa, line 94: Remove edit button in encryptable row, because the receive-recrypt only starts, when user clicks in the credentials. If he clicks edit before the receive-recrypt, all fields are null.

frontend/app/models/encryptable-transfer.js
```diff
@@ -0,0 +1,56 @@
...
+
+  async save() {
+    if (this.isDeleted) {
+      return super.save();
+    }
+    const targetUrl = `/api/encryptables_transfer`;
+    const receiverId = await this.receiver.get("id");
+    let opts;
+
+    if (this.description === undefined && this.file === undefined) {
```
mtnstar, line 18: extract to getter with meaningful naming of this condition

mtnstar (changes requested):
almost there ... try to make naming function even simpler !

frontend/app/models/encryptable-transfer.js
lkleisa: Check, if this is a good check, if not replace it and extract check to a getter

app/models/encryptable.rb
```diff
@@ -78,7 +82,7 @@ def encrypt_attr(attr, team_password)
       cleartext_value.presence &&
         Crypto::Symmetric::Aes256.encrypt(cleartext_value, team_password)
 
-    build_encrypted_data(attr, encrypted_value)
+    build_encrypted_data(attr, encrypted_value) unless encrypted_value.nil?
```
lkleisa, line 85: Make sure no nil value from custom attrs will be encrypted

app/utils/encryptable_transfer.rb
```diff
@@ -28,6 +33,75 @@ def receive(encryptable, private_key, personal_team_password)
...
+    regex_pattern = /\A#{Regexp.escape(encryptable_name)}(\(\d+\))?\z/
+    existing_encryptable_names.select do |name|
+      current_suffix = File.extname(name)
+      current_encryptable_name = File.basename(name, '.*')
+      current_encryptable_name.match?(regex_pattern) && current_suffix == encryptable_suffix
+    end
+  end
+
+  def remove_optional_climbs(encryptable_name)
+    regex_pattern = /\A.*\(\d+\)\z/
```
mtnstar, line 96: maybe also a better name for this ... what does this regex describe?

frontend/app/components/encryptable/show.js
```diff
@@ -21,6 +21,12 @@ export default class ShowComponent extends Component {
   @tracked
   isFile = this.args.encryptable.isFile;
 
+  @tracked
+  isCredentialSharing = false;
+
+  @tracked isTransferredCredentials =
+    this.args.encryptable.sender_name !== null;
```
mtnstar, line 28: is this really reliable? doesn't it also apply for transferred file entries?

`hasSender` might also be a possible name

frontend/app/models/encryptable-transfer.js
```diff
@@ -0,0 +1,61 @@
...
+  @attr file;
+  @service router;
+  @service fetchService;
+
+  async save() {
+    if (this.isDeleted) {
+      return super.save();
+    }
+    const targetUrl = `/api/encryptables_transfer`;
+    const receiverId = await this.receiver.get("id");
```
mtnstar, line 15:
```suggestion
    const receiverId = await this.receiver.id;
```

frontend/app/templates/components/encryptable/row.hbs
```diff
@@ -40,58 +40,75 @@
...
-                                         @visibleByDefault={{false}}/>
-          </div>
-        {{/if}}
-        {{#if (is-included @encryptable.usedEncryptedDataAttrs "custom_attr")}}
-          <div class="encryptable-container d-none d-md-block d-lg-block d-xl-block mr-0 ml-0 mr-md-2 ml-md-5 mr-lg-2 ml-lg-5 mr-xl-2 ml-xl-2">
-            <Encryptable::AttributeField @attribute="customAttr" @row={{true}} @encryptable={{@encryptable}}
-                                         @visibleByDefault={{false}}/>
-          </div>
-        {{/if}}
+      {{#if this.isTransferredCredentials }}
```
mtnstar, line 52:
```suggestion
      {{#if this.wasSent }}
```

```suggestion
      {{#if this.hasSender }}
```

mtnstar: does it make sense to show this information outside inbox folder?

so if someone moves the received encryptable to another folder ... maybe we can just hide it?

mtnstar:
image

would it make sense to call this title 'Share credentials with` ?

and then the hint:

instead of `Your credentials will be sent encrypted to the inbox folder from the selected receiver.`

`A copy of this credential will be sent to the receiver's inbox`

mtnstar:
at encryptable#show ...

image

how about: `Was sent by Alice Schindler (blabla) 22.06.2023 10:45`

spec/unit/utils/encryptable_transfer_spec.rb
```diff
@@ -57,4 +80,188 @@
...
+
+        allow(receiver).to receive_message_chain(:inbox_folder, :encryptables,
+                                                 :pluck).and_return(@existing_names)
+        destination_name = encryptable_transfer.send(:encryptable_destination_name, encryptable,
+                                                     receiver)
+
+        expect(destination_name).to eq('invoice.pdf')
+      end
+
+      it 'Keeps "invoice.pdf" if "invoice other.pdf" exist' do
```
mtnstar, line 160: how about the case ...

`blabliblu invoice.pdf` ?

frontend/app/templates/components/encryptable/row.hbs
lkleisa: In my view, the renaming is not needed, because the variable name says, what value is saved in it?!

frontend/app/components/encryptable/show.js
lkleisa: In my view, the renaming is not needed, because the variable name says, what value is saved in it?!

mtnstar (approved):
done :partying_face: 

we wait to merge this into master after we've done a production release 5.0 and fixed the encoding issue with transferred file.

PR #711, opened 2023-06-05
Feature/704; Allow csp font locally

lkleisa:


frontend/app/styles/secret-input.scss
```diff
@@ -1,15 +1,13 @@
 //  font from: https://github.com/noppa/text-security
 
 @font-face {
-  font-family: "text-security-disc";
-  src: url("/text-security-disc-compat.eot");
-  src: url("/text-security-disc-compat.eot?#iefix") format("embedded-opentype"),
-    url("/text-security-disc.woff2") format("woff2"),
-    url("/text-security-disc-compat.ttf") format("truetype");
+  font-family: text-security-disc;
+  src: url("https://raw.githubusercontent.com/noppa/text-security/master/dist/text-security-disc.woff");
```
mtnstar, line 5: don't load the font from an external source ... try to serve a copy from cryptopus assets

mtnstar (changes requested):
don't load the font from an external source ... try to serve a copy from cryptopus assets

njaeggi:
I have also taken another look, tried different methods to allow cors or set the Access-Control-Allow-Origin header, nothing really worked, I also tried to just change the URL to localhost:3000 since we have the assets loaded there too. This also didn't work

frontend/app/styles/secret-input.scss
```diff
@@ -2,10 +2,10 @@
 
 @font-face {
   font-family: "text-security-disc";
-  src: url("/text-security-disc-compat.eot");
-  src: url("/text-security-disc-compat.eot?#iefix") format("embedded-opentype"),
-    url("/text-security-disc.woff2") format("woff2"),
-    url("/text-security-disc-compat.ttf") format("truetype");
+  src: url("http://localhost:3000/text-security-disc-compat.eot");
```
TheWalkingLeek, line 5: I don't know much about CORS but I can't imagine this working on anything but a local development setup.

It works locally, but I hardly doubt that it will once it is deployed since we're just hardcoding the local url into this... Maybe someone else has to look into this :sweat_smile:

TheWalkingLeek (changes requested):

PR #715, opened 2023-06-12
fix styling of received file row

njaeggi:


mtnstar:
image

1. download icon does not have same styling as the other, and a strange _ when hovering
2. line-break needed for sender / send timestamp info?

PR #717, opened 2023-06-13
Remove deprecations

kcinay055679:


njaeggi:
When you encounter the Watchman Poison: inotify_add_watch error, you can increase the number of watches with this command:
 `echo 999999 | sudo tee -a /proc/sys/fs/inotify/max_user_watches && echo 999999 | sudo tee -a /proc/sys/fs/inotify/max_queued_events && echo 999999 | sudo tee -a /proc/sys/fs/inotify/max_user_instances && watchman shutdown-server && sudo sysctl -p`
 
 you can also jsut look up your max_user_watches and instances with this command:
 
`sudo sysctl fs.inotify.max_user_watches fs.inotify.max_queued_events fs.inotify.max_user_instances`

njaeggi:
I tried fixing the markdown error, but wasn't able to fix it yet. I tried it with different methods and also the axios dependency. Notghin really worked. I also downgraded bootstrap on this branch, since it didn't affect any deprecation warnings and just "destroyed" the UI because of ember-bootstrap.

When the bootstrap upgrade on this branch had any reason we can upgrade it back again. 

Also I removed the passwordStrength service, because it isn't existing anymore, I replaced it with https://github.com/elwayman02/ember-cli-password-strength which basically does the exact same thing, but different.

Also had a lot of problems with the versions and stuff, I didn't really know why some dependencies were upgraded on this branch.

As far as I know this is the stuff left todo:
- [ ] Fix changelog markdown loading
  - [ ] fix prettier for changelog component
  - [ ] integration test should run without any needed changes when markdown works again
- [ ] Remove deprecation warning for bowerDependencies or "bower-config" it is used by ember-cli (will probably be gone when upgrading ember on other branch)
- [ ] Remove deprecation warning for getWithDefault (appears on browser console when opening encryptable form)

Everything else should be running and working just fine, as far as I tested (Important before merging is clicking through the whole application)

Also weird is that the ember Inspector says the application runs on ember  3.24.7

mtnstar:
I moved changelog.md file to rails public folder and it's rendered now by markdown-it.

needs some styling ... but we're getting there!

frontend/app/initializers/env-settings.js
```diff
@@ -3,33 +3,36 @@ import ENV from "../config/environment";
...
-        ENV.lastLoginMessage = envSettings.last_login_message;
-        ENV.geoIP = envSettings.geo_ip;
-        ENV.appVersion = envSettings.version;
-        ENV.CSRFToken = envSettings.csrf_token;
-        ENV.authProvider = envSettings.auth_provider;
-        ENV.fallbackInfo = envSettings.fallback_info;
-      }
-    });
-    /* eslint-enable no-undef  */
+    const req = new XMLHttpRequest();
```
mtnstar, line 6: maybe use more modern `fetch` instead of `XMLHttpRequest()`

frontend/config/optional-features.json
```diff
@@ -1,3 +1,3 @@
 {
-  "jquery-integration": true
+  "jquery-integration": false
```
mtnstar, line 2: nice, we're getting rid of jquery! :partying_face:

mtnstar (approved):

PR #725, opened 2023-06-21
Feature/655 ember upgrade 4.x

kcinay055679:
Prioritise the [ember 3.28 branch](https://github.com/puzzle/cryptopus/tree/feature/655-ember-upgrade-3-28)


mtnstar:
for the frontend upgrade, the following things came up my mind:

- make sure you upgraded all dependencies/ember addons to the most recent version
- maybe you could also upgrade ember bootstrap
- temporarily remove all components from the application, does the frontend start? then re-add for example admin/users, make sure it's working ... and so on
- create a new ember app with the same version inside the docker container
- try to use the official nodejs container instead of the given ember container
  - https://hub.docker.com/_/node

spec/system/update_password_modal_system_spec.rb
```diff
@@ -73,7 +73,7 @@
 
   def fill_modal(password_attrs)
     within('div.modal-body') do
-      expect(page).to have_css('input', count: 3)
+      expect(page).to have_css('input')
```
mtnstar, line 76: why did that change from 3 to 1 ?

frontend/package.json
```diff
@@ -24,48 +24,50 @@
     "@babel/core": "^7.0.0",
```
mtnstar, line 24: did we upgrade all the packages? so all dependencies that we can update are in the most recent version?

mtnstar (changes requested):
very sexy ... also the whole dev env runs much smoother !

some small things to check

frontend/app/components/side-nav-bar.js
```diff
@@ -38,9 +38,9 @@ export default class SideNavBar extends Component {
   @action
   setSelectedTeam(team) {
     let alreadyTheSame = this.navService.selectedTeam === team;
-
-    if (alreadyTheSame) this.collapsed = !this.collapsed;
-    else {
+    if (alreadyTheSame) {
```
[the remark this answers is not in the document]
MarcEgliP: Changed this to "alreadyActiveTeam"

frontend/app/routes/application.js
```diff
@@ -3,11 +3,18 @@ import { inject as service } from "@ember/service";
...
+export default class ApplicationRoute extends Route {
   @service notify;
   @service intl;
+  @service logoutTimerService;
+  @service store;
 
-  beforeModel() {
+  async beforeModel() {
+    if(ENV.environment !== "test") {
+      const response = await fetch("/api/env_settings");
```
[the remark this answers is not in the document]
MarcEgliP: Not possible, it has to be done in the beforeModel, otherwise it won't work

spec/system/update_password_modal_system_spec.rb
MarcEgliP: I changed that back to assert  3 input fields.

frontend/package.json
MarcEgliP: Yes, now I've upgraded all the packages to their latest versions, except those: https://codimd.puzzle.ch/tGhXgK1kTKimo0Cv473FJA#

frontend/app/services/env-settings-service.js
```diff
@@ -0,0 +1,33 @@
+import Service from "@ember/service";
+import ENV from "../config/environment";
+
+export default class EnvSettingsService extends Service {
+
+  async handleEnvironmentSettings() {
```
mtnstar, line 6: rename method to fetchSettings

frontend/app/services/env-settings-service.js
```diff
@@ -0,0 +1,33 @@
...
+    if(ENV.environment !== "test") {
```
mtnstar, line 7: move this check to route ... no need to add this inside this service

PR #727, opened 2023-06-26
Feature/712/file without credentials

njaeggi:


Robin481 (changes requested):
1) Make the whole file field clickable and not just the left half. 
2) Allow uploading the same file in different folders 
3) When selecting to upload a file in a existing folder via the plus button the team from which you pressed the plus button should automatically be selected in the folder dropdown
4) If no folder is selected the form breaks quite badly. It should display a proper error

njaeggi:
I implemented the last three suggestions, the first one should be done in another ticket, I can't find the issue, I thought that a ticket like that already existed, we have no click action on the whole row on any type of encryptable(file or credential). This should be changed on both type of rows. Should I open a ticket, I think the issue was accidentally closed during issue cleanup...

app/controllers/api/encryptables_controller.rb
```diff
@@ -68,7 +68,7 @@ def model_class
   end
 
   def define_model_class
-    if credential_id.present?
+    if credential_id.present? || params[:folder_id].present?
```
mtnstar, line 71: interesting ... why does this work with the folder_id?
doesn't have a credential also have a folder_id ?

I would say this check is rather a bit shaky .. is there a more solid way to determine the model class?

app/controllers/api/encryptables_controller.rb
```diff
@@ -126,7 +130,7 @@ def permitted_attrs
     permitted_attrs = self.class.permitted_attrs.deep_dup
 
     if model_class == Encryptable::File
-      permitted_attrs + [:filename, :credentials_id, :file]
+      permitted_attrs + [:filename, :credentials_id, :folder_id, :file]
```
mtnstar, line 133: looks like folder_id is available for both model classes ... maybe add to default?

app/controllers/api/encryptables_controller.rb
```diff
@@ -165,19 +169,29 @@ def fetch_encryptable_files
...
+             new_file(folder, params[:description], filename)
+           end
     file.content_type = params[:file].content_type
     file.cleartext_file = params[:file].read
 
     instance_variable_set(:"@#{ivar_name}", file)
   end
 
-  def new_file(parent_encryptable, description, name)
+  def new_attachment(parent_encryptable, description, name)
```
mtnstar, line 183: I would rather find a way to define the correct model params and use dry crud functionality for creating the file entry.

app/models/encryptable/file.rb
```diff
@@ -10,6 +10,7 @@ class Encryptable::File < Encryptable
              foreign_key: :credential_id
 
   validates :name, uniqueness: { scope: :credential_id }, if: :credential_id
+  validates :name, uniqueness: { scope: :folder_id }, if: :folder_id
```
mtnstar, line 13: nice :+1:

public/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## Version 5.3
+- Introduce uploading files without adding them to credentials
```
mtnstar, line 2: - Introduce encryptable files for folders

mtnstar (changes requested):
looks pretty good already! :partying_face: 

check the encryptables controller and make it more D.R.Y by using default functionality of dry-crud: https://github.com/codez/dry_crud

app/controllers/api/encryptables_controller.rb
kcinay055679: No, the attachment just have a credential_id not a folder id. Since we don't have any useful params the check via folderId is the best solution.

app/controllers/api/encryptables_controller.rb
```diff
@@ -165,23 +169,18 @@ def fetch_encryptable_files
...
-    Encryptable::File.new(encryptable_credential: parent_encryptable,
-                          description: description,
-                          name: name)
-  end
-
   def credential_id
     return params[:id] if params[:id].present?
 
-    params[:credential_id]
+    params[:credential_id] == 'null' ? nil : params[:credential_id]
```
TheWalkingLeek, line 183: dang does the ember really send `"null"` for a null value? I feel like that shouldn't happen
If it really does, maybe it would make sense to make a method `nil_param?` or something for the ApplicationController and then use it here

TheWalkingLeek (changes requested):
`upload_file` & `submit` buttons arent translated in every language
Also I'm getting a browser console error when closing the modal

TheWalkingLeek (approved):

PR #735, opened 2023-06-29
no encryptable bug in folder

njaeggi:
I noticed this while working on file without credentials, the fix was pretty easy and should work as intended. I also switched the translations, to credential or files, since we implement file creation directly and file sharing also enables the possibility to have folder attached files and not credentials


lkleisa:
I would approve this masterpiece :1st_place_medal:

Robin481 (approved):

PR #737, opened 2023-07-03
Feature/726/ccli improvement

njaeggi:


MarcEgliP (approved):
Seems good to me!

lkleisa:
For me, the implementation didn't work with the copy buttons. I figured out, that the copy function in our clipboard service is soooo long, so I replaced our own method to copy something to clipboard with a one-liner of JavaScript, which is working on all browsers: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

PR #759, opened 2024-01-29
Bug/757-fix-encryptable-search

MarcEgliP:
What was made:
- Search has been fixed (When searching for an encryptable it will return only the searched one)
- A e2e test has been added to verify the above described behaviour
- One Unit Test has been ignored because it requires ordering the data after it has been selected 
=> We will fix this in another Ticket: #760

### DIFF
--- a/spec/controllers/api/teams_controller_spec.rb
+++ b/spec/controllers/api/teams_controller_spec.rb
@@ -180,7 +180,10 @@ describe Api::TeamsController do
       expect(folder_relationships_length).to be(1)
     end
 
-    it 'returns encryptable files for team_id, in order from created_at' do
+    ## TODO: Activate test after order logic has been implemented again
+    # Has been ignored because order logic in serializer had to be removed to make search work
+    # Ticket: https://github.com/orgs/puzzle/projects/5/views/1?pane=issue&itemId=51495477
+    xit 'returns encryptable files for team_id, in order from created_at' do
       inbox_folder_receiver = alice.inbox_folder
       personal_team_alice = teams(:personal_team_alice)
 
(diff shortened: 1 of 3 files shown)


spec/system/search_system_spec.rb
```diff
@@ -61,6 +61,23 @@
     end
   end
 
+  it 'finds matching encryptable' do
+    login_as_user(:bob)
+    visit('/')
+
+    encryptable1 = encryptables(:credentials2)
```
mtnstar, line 68: maybe create some more encryptables in same folder to make sure those are not rendered?
you might use fabricator for this
you then can add expectations that those are not rendered at all.

spec/controllers/api/teams_controller_spec.rb
```diff
@@ -180,7 +180,10 @@
       expect(folder_relationships_length).to be(1)
     end
 
-    it 'returns encryptable files for team_id, in order from created_at' do
+    ## TODO: Activate test after order logic has been implemented again
+    # Has been ignored because order logic in serializer had to be removed to make search work
+    # Ticket: https://github.com/orgs/puzzle/projects/5/views/1?pane=issue&itemId=51495477
```
mtnstar, line 185: use direkt github link without project board ....
this is not the follow up ticket for fixing this, is it?
better just link the follow up ticket, and inside the new ticket reference to the old one

mtnstar (changes requested):

spec/controllers/api/teams_controller_spec.rb
MarcEgliP: I changed the link to navigate directly to the follow up issue and referenced this ticket in the follow up story.

mtnstar: how about a shorter comment:
TODO: Re-enable after fixing encryptables ordering, see #760

spec/system/search_system_spec.rb
MarcEgliP: I added two more credentials to the same folder via fabricator.

